PR #25, opened 2017-05-30
Parse JUnit test results

cpolcyn:


cpolcyn:
If no test results found, return instead of raising exception.

jlantz:
I've determined that the browser tests can output JUnit but as multiple files under a directory.  Since other testing commands wrapped via the Command task might output JUnit to other locations too, the best way to support this seems to be to allow configuration of the JUnit file(s) location on each Plan.

Add a new field, Plan.junit_path, that supports wildcards such as `junit/*.xml` and have the JUnit loader at the end of builds for the plan look for JUnit files in the specified location if provided.  Otherwise, just try to find test_results.xml as cci would normally output it.

cpolcyn:
Order of test results search:

1. Specified file or directory of files (via Path.junit_path)
2. Debug test info (test_results.json)
3. Apex test results (test_results.xml)

PR #29, opened 2017-06-13
limit model log db writes to 1/sec

### LINKED ISSUE #15: Batch write log messages to db
Currently, the log handler that writes flow logs to the database is writing each line emitted to the logger as its own commit.  This starts to show signs of strain under load (~90 concurrent builds) both in postgresql and in django-watson (full text search engine) indexing.

The log handler should cache log messages and write batched log messages to the db no more than 1 commit per second.

cpolcyn:
Fixes #15

### DIFF
--- a/mrbelvedereci/build/models.py
+++ b/mrbelvedereci/build/models.py
@@ -91,6 +91,10 @@ class Build(models.Model):
         url = '{}{}'.format(settings.SITE_URL, self.get_absolute_url())
         return url
 
+    def flush_log(self):
+        for handler in self.logger.handlers:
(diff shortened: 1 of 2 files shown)


mrbelvedereci/build/models.py
```diff
@@ -117,6 +117,7 @@ def run(self):
             self.time_end = timezone.now()
             self.save()
             self.delete_build_dir()
+            self.logger.handler.stream.flush(force=True)
```
cpolcyn: RE: self.logger.handler it seems it would be invalid but I confirmed this worked in my testing…

PR #37, opened 2017-07-12
throw ServiceNotConfigured instead of Service.DoesNotExist

### LINKED ISSUE #36: throw ServiceNotConfigured instead of Service.DoesNotExist
if mbci's keychain get_service fails, the standard django DoesNotExist exception occurs.

we should catch and raise ServiceNotConfigured, since cumulusci.core.config expects that (for example, to check if sentry is enabled)

cdcarter:
closes #36 (throw ServiceNotConfigured instead of Service.DoesNotExist) which occurs if sentry isn't configured locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@
 # proportion of contributors will probably not be using SublimeText
 # *.sublime-project
 
+# vscode
+.vscode/*
+
 # sftp configuration file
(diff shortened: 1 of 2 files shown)


cdcarter:
@cpolcyn - no rush on a review here, but would be appreciated.

PR #39, opened 2017-07-21
test_results.json may contain actual integers

cdcarter:
debug log based limits usage values are in a string, because they may have the "CLOSE TO LIMIT" text.

however, limit usage values from the tooling API will be a true int. EAFP - try and strip that text but don't worry about it too much.


cdcarter:
one last little PR @cpolcyn, and then I'll stop bugging you for a day. :)

PR #52, opened 2017-08-04
Refresh OAuth token before logging in to org

cpolcyn:
It would be nice to use MrbelvedereProjectKeychain to get the connected app info and not duplicate the code here, but it requires a project_config which in turn requires us to be in a build directory to retrieve the project_config. Since the login action could occur after the build dir is long gone, I think we have to do it this way.


cpolcyn:
Just added one more commit to move the get_connected_app stuff into a new file that both the view and keychain use.

PR #65, opened 2017-08-09
Feature/testcompare

cdcarter:
you need to know how to get there, adding a way to select builds to compare is coming. But, compare the test limits usage from two buildflows:

screen shot 2017-08-09 at 2 25 27 pm

### DIFF
--- a/mrbelvedereci/testresults/models.py
+++ b/mrbelvedereci/testresults/models.py
@@ -49,15 +49,11 @@ class TestResultManager(models.Manager):
                     results[cls][method] = {}
     
                 for limit in result.get_limit_types():
-                    limit = limit + '_used'
-                    test_limit = 'test_' + limit
+                    test_limit = 'test_{}_used'.format( limit )
     
-                    if limit not in results[cls][method]:
-                        results[cls][method][limit] = OrderedDict()
                     if test_limit not in results[cls][method]:
                         results[cls][method][test_limit] = OrderedDict()
     
-                    results[cls][method][limit][build_flow.id] = getattr(result, limit)
                     results[cls][method][test_limit][build_flow.id] = getattr(result, test_limit)
     
         diff = OrderedDict()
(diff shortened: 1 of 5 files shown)


mrbelvedereci/testresults/views.py
```diff
@@ -7,7 +7,7 @@
 from mrbelvedereci.build.models import Build
 from mrbelvedereci.build.models import BuildFlow
 from mrbelvedereci.build.utils import paginate
-from mrbelvedereci.testresults.models import TestMethod
+from mrbelvedereci.testresults.models import TestMethod, TestResult
```
cpolcyn: Please do this as a new import line for consistency

mrbelvedereci/testresults/models.py
```diff
@@ -49,15 +49,11 @@ def compare_results(self, build_flows):
...
-                    limit = limit + '_used'
-                    test_limit = 'test_' + limit
+                    test_limit = 'test_{}_used'.format( limit )
     
-                    if limit not in results[cls][method]:
-                        results[cls][method][limit] = OrderedDict()
                     if test_limit not in results[cls][method]:
                         results[cls][method][test_limit] = OrderedDict()
     
-                    results[cls][method][limit][build_flow.id] = getattr(result, limit)
```
cpolcyn: What's the reason for removing this?

cdcarter: the tooling api test runner means we only care about the `test_` limits (the transactions between Test.startTest() and Test.startTest()).

so, the non `test_` limits will be None on both sides of the comparison.

mrbelvedereci/testresults/views.py
```diff
@@ -8,6 +8,7 @@
 from mrbelvedereci.build.models import BuildFlow
 from mrbelvedereci.build.utils import paginate
 from mrbelvedereci.testresults.models import TestMethod
+from mrbelvedereci.testresults.model import TestResult
```
cpolcyn: typo: should be "models"

PR #72, opened 2017-08-11
Show Rebuild button only when build status is error or fail

### LINKED ISSUE #50: Rebuilding stuck waiting on previous build
Rebuilds against persistent orgs should set previous rebuilds and original build to non-running status and release org lock.

cpolcyn:
Fixes #50


cpolcyn:
For build in waiting/queued/running we would want to first cancel the build (see #5) which should set status to error, then rebuild would be possible. Until we can cancel a build, this will prevent some problems as described in #5. Builds stuck in waiting/queued/running should be manually set to error by an admin.

PR #77, opened 2017-08-14
Audit manual builds

### LINKED ISSUE #73: Manual build should record and show user
For manual builds we should record which user started the build so it can be used in e.g. User column of rebuilds list view.

cdcarter:
Adds two fields to Build:
* user - foreignkey relationship to User for builds initiated by a user
* build_type - auto, manual, scheduled, legacy (default, for past records in the migration)

When a manual build is created, the build user field is set to the running user.
~~When a manual build is created, a CREATED entry is added to the admin log, with the running user. (Is this actually necessary? Does it make sense?)~~

The build user is displayed on the "Rebuilds" tab of the build detail when the original build was created by a user. 

closes #73 - Manual build should record and show user


mrbelvedereci/plan/handlers.py
```diff
@@ -2,8 +2,31 @@
...
 from django.dispatch import receiver
+from django.contrib.admin.models import LogEntry, ADDITION
+from django.contrib.contenttypes.models import ContentType
 from github3 import login
 from mrbelvedereci.plan.models import Plan
+from mrbelvedereci.build.models import Build
+
+
+@receiver(post_save, sender=Build)
+def log_manual_build(sender, **kwargs):
```
cpolcyn: Probably not necessary since we are already getting this info in the objects. Any added benefit?

cdcarter: It was really just to play around with LogEntry. Since the admin log only has three action types - Created, Changed, and Deleted, it doesn't have a very expressive or useful UI. there's also no out of the box Lists of LogEntry in the admin interface. 

I don't think it's a great solution for any of our audit logging needs, really. I'll remove it from the PR.

PR #78, opened 2017-08-14
Select a build to compare to

cdcarter:
when viewing a build's [test limits usage](https://mrbelvedereci.herokuapp.com/tests/6187/ci_feature) this PR adds a Compare button, that displays a list of recent builds to compare it to.

test compare

adds a new page & view to list compare options, a helper to lookup BuildFlow based on the tests/ url patterns, and uses django_filters to filter the queryset *and* generate a form for the client.


mrbelvedereci/testresults/urls.py
```diff
@@ -5,6 +5,11 @@
 
 urlpatterns = [
     url(
+        r'^(?P<build_id>\d+)/(?P<flow>.*)/compare_to',
```
cpolcyn: I'd like this changed to `compare-to` but is just my preference.

mrbelvedereci/testresults/utils.py
```diff
@@ -0,0 +1,17 @@
...
+from mrbelvedereci.build.models import Build
+from mrbelvedereci.build.models import BuildFlow
+
+
+def find_buildflow(request, build_id, flow):
+    """ given a build_id and flow name, find a single BuildFlow (ala tests/ urls patterns). """
+    build = get_object_or_404(Build, id=build_id)
+
+    if not build.plan.public and not request.user.is_staff:
+        raise Http404()
```
cpolcyn: Should be 403 Forbidden

requirements/base.txt
```diff
@@ -11,6 +11,7 @@ whitenoise==3.2.2
 # Forms
 django-braces==1.10.0
 django-crispy-forms==1.6.1
+django_filter==1.0.4
```
cpolcyn: Why is this needed? Can't this be done simply using URL query params passed in the request and use Django object query?

mrbelvedereci/testresults/utils.py
```diff
@@ -0,0 +1,17 @@
+from django.shortcuts import get_object_or_404
+from mrbelvedereci.build.models import Build
+from mrbelvedereci.build.models import BuildFlow
+
+
+def find_buildflow(request, build_id, flow):
+    """ given a build_id and flow name, find a single BuildFlow (ala tests/ urls patterns). """
+    build = get_object_or_404(Build, id=build_id)
+
+    if not build.plan.public and not request.user.is_staff:
```
cpolcyn: `is_staff` limits to only admins – maybe you want `is_authenticated`?

cdcarter: <redacted_adminsite_url>/users/user/?is_staff__exact=1 - `is_staff` is internal users, `is_superuser` is admins. 

an auth helper for builds in general would be nice, and I know we have a ticket for using github access...

requirements/base.txt
cdcarter: the library (and filter subclass) is also generating the form on the page (search fields for build name and plan name).

it's not strictly necessary, but it meant I didn't have to build a form. would it be worth it for me to try the other method and compare?

mrbelvedereci/testresults/utils.py
```diff
@@ -8,7 +9,7 @@ def find_buildflow(request, build_id, flow):
     build = get_object_or_404(Build, id=build_id)
 
     if not build.plan.public and not request.user.is_staff:
-        raise Http404()
+        raise Http403()
```
cpolcyn: `Http403` doesn't exist, should be `HttpResponseForbidden`. Sorry if my previous comment was unclear!

cdcarter: No, that's my bad. my server is complaining and everything.

requirements/base.txt
cpolcyn: The in-page form is cool and seems worth it for that. Thanks for clarifying!

PR #83, opened 2017-08-15
add notification if developing locally

cdcarter:
adds a notification toast with the hostname if developing locally in debug mode

also sets a base title in the layout of mrbelvedereci

### DIFF
--- a/mrbelvedereci/templates/layout_full.html
+++ b/mrbelvedereci/templates/layout_full.html
@@ -18,7 +18,7 @@
       <span class="slds-assistive-text">Info</span>
 
       <div class="slds-notify__content">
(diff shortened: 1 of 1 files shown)


mrbelvedereci/templates/layout_full.html
```diff
@@ -2,16 +2,31 @@
...
-  <div class="slds-global-header slds-grid slds-grid--align-spread">
+
+{% block local_header %}
+{% if debug %}
+<div class="slds-notify_container">
+    <div class="slds-notify slds-notify--toast" role="alert">
+      <span class="slds-assistive-text">Info</span>
+
+      <div class="slds-notify__content">
+        <h2 class="slds-text-heading--small">DEVELOPING ON {{request.META.HTTP_HOST}}</h2>
```
cpolcyn: I think the convention is to have inner spaces for these vars like {{ x }}

PR #84, opened 2017-08-15
Use timezone.now instead of datetime.utcnow

cpolcyn:
Makes the time objects timezone aware

### DIFF
--- a/mrbelvedereci/cumulusci/logger.py
+++ b/mrbelvedereci/cumulusci/logger.py
@@ -6,6 +6,7 @@ import datetime
 import logging
 
 import coloredlogs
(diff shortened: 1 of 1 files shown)


cdcarter:
which timezone does this use? the timezone of the django user who kicked off the build? the timezone of the build worker task process?

cpolcyn:
It's timezone aware, so under the hood it's always UTC and can be adapted to other timezones as needed based on browser location.

cpolcyn:
timezone.now is used elsewhere in the code so this is really just bringing these stragglers in line.

cdcarter:
ah, timezone *aware* - cool!

PR #88, opened 2017-08-17
run migrate in a non interactive env

cdcarter:


cdcarter:
We'll need to clean up the stale content type after the build is done, but that's fine and not urgent.

PR #92, opened 2017-08-17
local development - no github status updates

cdcarter:
Adds env var and setting `GITHUB_STATUS_UPDATES_ENABLED`

defaults to True, defaults to False in local.py. 

When false, RQ jobs to update the github commit status are not created, so as not to pollute github.


config/settings/common.py
```diff
@@ -290,3 +290,5 @@
 CONNECTED_APP_CLIENT_ID = None
 CONNECTED_APP_CLIENT_SECRET = None
 CONNECTED_APP_CALLBACK_URL = None
+
+GITHUB_STATUS_UPDATES_ENABLED = env.bool('GITHUB_STATUS_UPDATES_ENABLED', True)
```
cpolcyn: Would it make more sense to default=False and set =True in production.py?

mrbelvedereci/build/handlers.py
```diff
@@ -16,8 +17,9 @@ def queue_build(sender, **kwargs):
         return
 
     # Queue the pending status task
-    res_status = set_github_status.delay(build.id)
-    build.task_id_status_start = res_status.id
+    if settings.GITHUB_STATUS_UPDATES_ENABLED:
+        res_status = set_github_status.delay(build.id) 
+        build.task_id_status_start = res_status.id
```
cpolcyn: Are `task_id_status_start` and `task_id_status_end` used anywhere? I looked around and didn't see any. Just curious.

cdcarter: nope, they are just for managing the child task, which won't be created in this case.

config/settings/common.py
cdcarter: I was thinking about this. I think this way makes sense, because in TEST we also want to go down the code path, and instead mock out the task and subsequent api call.

PR #112, opened 2017-09-05
Add button for superusers to lock/unlock orgs

cpolcyn:
If user is superuser, and org is persistent, show a button to lock/unlock the org manually with no timeout.

### DIFF
--- a/mrbelvedereci/build/tasks.py
+++ b/mrbelvedereci/build/tasks.py
@@ -98,9 +98,11 @@ def check_queued_build(build_id):
 
     else:
         # For persistent orgs, use the cache to lock the org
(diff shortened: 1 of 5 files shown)


cdcarter:
should we require POST and CSRF token to these endpoints?

cpolcyn:
Don't want to require POST since this is just coming from a href in the template. Regarding CSRF token I thought it'd be sufficient to require `user.is_superuser`.

cpolcyn:
I see, even though the is_superuser restriction will protect us, it is kind of an abuse of the GET. I'll make the changes to do a POST. We can use this button to redirect to a is_superuser-only view that asks for confirmation and does a POST.

PR #119, opened 2017-09-12
Manually lock/unlock org

cpolcyn:
Continuation of https://github.com/SalesforceFoundation/mrbelvedereci/pull/112

Now the (super)user is sent to a confirmation form. This ensures we send a POST to modify the lock state, and CSRF protection is taken care of by crispy-forms.

### DIFF
--- a/mrbelvedereci/build/tasks.py
+++ b/mrbelvedereci/build/tasks.py
@@ -98,9 +98,11 @@ def check_queued_build(build_id):
 
     else:
         # For persistent orgs, use the cache to lock the org
-        lock_id = 'mrbelvedereci-org-lock-{}'.format(org.id)
-        status = cache.add(lock_id, 'build-{}'.format(build_id),
-                           timeout=BUILD_TIMEOUT)
+        status = cache.add(
+            org.lock_id,
+            'build-{}'.format(build_id),
+            timeout=BUILD_TIMEOUT,
+        )
 
         if status is True:
             # Lock successful, run the build
--- a/mrbelvedereci/cumulusci/models.py
+++ b/mrbelvedereci/cumulusci/models.py
@@ -36,6 +36,7 @@ class Org(models.Model):
         if not self.scratch:
             return u'mrbelvedereci-org-lock-{}'.format(self.id)
 
+    @property
     def is_locked(self):
         if not self.scratch:
             return True if cache.get(self.lock_id) else False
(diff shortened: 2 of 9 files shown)


mrbelvedereci/cumulusci/models.py
```diff
@@ -29,7 +30,25 @@ def get_org_config(self):
...
 
         return OrgConfig(org_config)
-    
+
+    @property
+    def lock_id(self):
+        if not self.scratch:
+            return u'mrbelvedereci-org-lock-{}'.format(self.id)
+
+    def is_locked(self):
```
cdcarter: could be `@property` too

mrbelvedereci/cumulusci/views.py
```diff
@@ -24,6 +28,36 @@ def org_detail(request, org_id):
     } 
     return render(request, 'cumulusci/org_detail.html', context=context)
     
+def org_lock_unlock(request, org_id, action):
```
cdcarter: it took me a minute to realize this wasn't actually wired up directly to a urlconf - can you add a comment or do we have a naming convention we can use to make that quicker to notice? just a thought.

mrbelvedereci/build/tasks.py
```diff
@@ -98,9 +98,11 @@ def check_queued_build(build_id):
 
     else:
         # For persistent orgs, use the cache to lock the org
-        lock_id = 'mrbelvedereci-org-lock-{}'.format(org.id)
-        status = cache.add(lock_id, 'build-{}'.format(build_id),
-                           timeout=BUILD_TIMEOUT)
+        status = cache.add(
+            org.lock_id,
+            'build-{}'.format(build_id),
```
cdcarter: should we have a helper func for the build id? i know its just a simple str format but it seems like we may want to be able to ref that logic again

mrbelvedereci/cumulusci/views.py
cpolcyn: Sure, I'll have it start with underscore and add a comment to make it clear.

mrbelvedereci/build/tasks.py
cpolcyn: Right, this same string is created elsewhere so it makes sense to consolidate it somewhere. FYI this value is never used (yet).

cpolcyn: Actually this is the only place it's used, and the value is never read, so I'll leave it be until we have some need for it.

cpolcyn:
@cdcarter I addressed your comments (mostly). Ready for re-review!

PR #126, opened 2017-09-20
Fix URL query for status in build list

### LINKED ISSUE #76: Build query by status doesn't consider rebuild status
We can filter the build list view with query param like so:
https://mrbelvedereci.herokuapp.com/builds/?status=error

Due to the changes in https://github.com/SalesforceFoundation/mrbelvedereci/pull/71 this query now yields builds where build.status matches, but rebuild status could be different. It should use build.get_status() instead.

cpolcyn:
Fixes #76

### DIFF
--- a/mrbelvedereci/build/utils.py
+++ b/mrbelvedereci/build/utils.py
@@ -7,6 +7,7 @@ from django.apps import apps
 from django.core.paginator import EmptyPage
 from django.core.paginator import PageNotAnInteger
 from django.core.paginator import Paginator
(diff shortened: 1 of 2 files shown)


cpolcyn:
Performance is pretty bad on this, we need to do this at database level instead. Something like this (thanks @jlantz):

```python
Q(build.current_rebuild__is_null = True, status = 'failed') | Q(build.current_rebuild__is_null = False, build.current_rebuild.status = 'failed')
```

PR #127, opened 2017-09-21
Update the Gemfile to also require json

jlantz:
The cci tasks that run browser tests first load the Gemfile from the tests.  In theory this is to allow individual projects to define additional gem dependencies.  The problem is that this happens in dyno runtime rather than slug compilation.  As a result, if any dependency in the tree generated from the Gemfile requires compilation with gcc it will fail.

It appears the core http gem in ruby is now requiring json which it wasn't before.  That caused all browser test builds to fail.  This branch just adds the json gem to mrbelvedereci's Gemfile so it's installed during slug compilation.  I then ran `bundle install` and committed the resulting Gemfile.lock

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ gem 'page-object'
 gem 'rspec'
 gem 'cucumber'
 gem 'restforce'
+gem 'json'
 gem 'SFDO-API'
 gem 'pry', require: false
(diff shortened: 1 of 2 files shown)


jlantz:
I bypassed the review requirement on this to get the fix into production for all the browser tests failing overnight.

PR #146, opened 2017-10-24
REST API (and other enhancements/cleanup)

jlantz:
This branch uses Django REST Framework and Django Filter to build a complete Core API enabled REST API.  The api is available at `/api` and has a complete web interface.  All API views require an authenticated user with the Django `is_staff` role.

# API Endpoints
The following json response from a local dev instance's `/api` endpoint shows the models covered by the API with GET, POST, HEAD, and OPTION requests for each model:
{
    "branch": "http://localhost:8000/api/branch/",
    "builds": "http://localhost:8000/api/builds/",
    "build_flows": "http://localhost:8000/api/build_flows/",
    "orgs": "http://localhost:8000/api/orgs/",
    "plans": "http://localhost:8000/api/plans/",
    "rebuilds": "http://localhost:8000/api/rebuilds/",
    "repos": "http://localhost:8000/api/repos/",
    "scratch_orgs": "http://localhost:8000/api/scratch_orgs/",
    "services": "http://localhost:8000/api/services/"
}

# Django Filter
Django Filter provides an easy way to define filtering interfaces controlled through url params on your models.  The filters support filtering across filter object relationships, however the relationships do not come through in the auto-generated Core API schema.  A future enhancement would be to expand the Core API schema generation to include all the related options available.

# Token Authentication
For now, admin users need to use the admin to generate tokens for each user who wants to use a token.  This should be enhanced in a future PR to provide a way for users to create their own token, but that will require 

# Other Changes

- Removed dependency on AWS to make configuration deployment easier
- Wired up the new BuildFilter to the build_list view.  On the home page's build list, you can now filter using all the filter options configured in BuildFilter such as `?status=failed&org__scratch=0&repo__name=Cumulus`
- Added a link to admin for all `is_staff` users
- Fixed plural names for `Repository` and `Branch` models


config/settings/common.py
```diff
@@ -169,7 +172,7 @@
                 'django.template.context_processors.static',
                 'django.template.context_processors.tz',
                 'django.contrib.messages.context_processors.messages',
-                # Your stuff: custom template context processors go here
+                'mrbelvedereci.users.context_processors.admin_url',
```
cdcarter: why a context processor, and not just reverse it with ` {% url 'admin:index' %}`

mrbelvedereci/api/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
cdcarter: rm

mrbelvedereci/api/models.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import unicode_literals
+
+from django.db import models
+
+# Create your models here.
```
cdcarter: rm

mrbelvedereci/api/serializers/build.py
```diff
@@ -0,0 +1,105 @@
...
+    org_instance = ScratchOrgInstanceSerializer(read_only=True)
+    plan = PlanSerializer(read_only=True)
+    plan_id = serializers.PrimaryKeyRelatedField(
+        queryset=Plan.objects.all(),
+        source='plan',
+        write_only=True
+    )
+    repo = RepositorySerializer(read_only=True)
+    repo_id = serializers.PrimaryKeyRelatedField(
+        queryset=Repository.objects.all(),
```
cdcarter: this manager isn't scoped to running user right? could this leak the names of our private repos if we opened up the API?

mrbelvedereci/repository/models.py
```diff
@@ -15,6 +15,7 @@ class Repository(models.Model):
 
     class Meta:
         ordering = ['name','owner']
+        verbose_name_plural = 'repositories'
```
cdcarter: thank. you.

mrbelvedereci/templates/layout_full.html
```diff
@@ -102,6 +102,22 @@ <h3 class="slds-text-heading--medium">Notifications</h3>
         </div>
         </a>
       </li>
+      {% if request.user.is_staff %}
```
cdcarter: do we want to add the interactive API browser  as a button too?

Procfile
```diff
@@ -1,4 +1,3 @@
-release: python manage.py migrate --noinput
```
cpolcyn: Why was this removed? Pretty useful for deploying.

mrbelvedereci/build/filters.py
```diff
@@ -0,0 +1,92 @@
...
+        model = Build
+        fields = {
+            'commit': ['exact'],
+            'status': ['exact'],
+            'time_queue': ['gt','lt'],
+            'time_start': ['gt','lt'],
+            'time_end': ['gt','lt'],
+        }
+
+class BuildFilter(BuildRelatedFilter):
```
cpolcyn: Why are these done this way instead of defining directly in BuildFilter?

mrbelvedereci/templates/layout_full.html
```diff
@@ -14,7 +14,7 @@
 {% block local_header %}
 {% if debug %}
 <div class="slds-notify_container">
-    <div class="slds-notify slds-notify--toast" role="alert">
```
cpolcyn: Does this fix https://github.com/SalesforceFoundation/mrbelvedereci/issues/96?

mrbelvedereci/templates/layout_full.html
cpolcyn: I'd prefer to keep it clean and not have a button for API.

mrbelvedereci/templates/layout_full.html
```diff
@@ -102,6 +102,22 @@ <h3 class="slds-text-heading--medium">Notifications</h3>
...
+      <li class="slds-dropdown-trigger slds-dropdown-trigger--click slds-p-horizontal--large">
+        <a href="/{{ ADMIN_URL }}">
+        <div class="slds-media">
+          <div class="slds-media__figure">
+            {% autoescape off %}
+            <svg class="slds-button__icon slds-global-header__icon" aria-hidden="true">
+              <use xlink:href="/static/slds/assets/icons/utility-sprite/svg/symbols.svg#settings"></use>
+            </svg>
+            {% endautoescape %}
+            <span class="slds-assistive-text">Notifications</span>
```
cpolcyn: Should be "Admin"

Procfile
jlantz: I need to add it back.  I removed it because the release command was swallowing an error earlier in the Heroku deployment process which I've now resolved after debugging without the release process.

mrbelvedereci/build/filters.py
jlantz: In anticipation that we'll want to have an option for more restricted filters via a relationship than we'd want on the object itself directly.  The main views of each object would use the *Filter while any filter that needs to reference that object would reference its *RelatedFilter.  The wiring is in place for that now so all we need to do to allow more specific filtering on the object itself is override the *Filter class

mrbelvedereci/templates/layout_full.html
jlantz: Yes, there's still an overlay over part of the header but all the header links and the search box are still usable this way.

config/settings/common.py
jlantz: Good point!

Honestly, it's been a while since I used a context processor and that was the references I found when searching this use case online.  But, I was searching for how to get access to settings rather than the more generic: "How do a build a link to admin in a template?".  I'll change to the tag and remove the context processor.

In the future, I anticipate we could gain a lot from using context processors more though.  They're super easy.  This just isn't the right use case.

mrbelvedereci/api/serializers/build.py
jlantz: That's correct.  For now, the API is configured to allow full access to is_staff users.  We'll work on more granular permissions in a future PR, sometime very soon though.

mrbelvedereci/templates/layout_full.html
jlantz: I agree with @cpolcyn for now.  I want to get the more granular restrictions set up on the API so we can open it up more.  Ideally there'd even be anonymous access to the API for read on public info that would be throttled via DRF.

mrbelvedereci/api/admin.py
jlantz: removed

mrbelvedereci/api/models.py
jlantz: removed

mrbelvedereci/build/management/commands/metaci_scheduled_jobs.py
```diff
@@ -0,0 +1,23 @@
...
+            defaults={
+                'interval': 1,
+                'interval_unit': 'minutes',
+                'scheduled_time': timezone.now(),
+            }
+        )
+        if created:
+            self.stdout.write(self.style.SUCCESS('Created job check_waiting_builds'))
+        else:
+            self.stdout.write(self.style.SUCCESS('Scheduled job check_waiting_builds already exists'))
```
cdcarter: if it already exists, can you render (at the very least) the ID and that its enabled/disabled?

mrbelvedereci/hirefire/views.py
```diff
@@ -25,6 +25,8 @@ def info(request, token):
     """
     Return the HireFire json data needed to scale worker dynos
     """
+    if not settings.HIREFIRE_TOKEN:
+        return HttpResponseForbidden('Hirefire not configured.  Set the HIREFIRE_TOKEN environment variable on the app to use Hirefire for dyno scaling')
```
cdcarter: throw a 500, not a 400?

mrbelvedereci/users/management/commands/usertoken.py
```diff
@@ -0,0 +1,17 @@
+from django.core.management.base import BaseCommand, CommandError
```
cdcarter: wait. i already reviewed this. is this command in here twice?

jlantz: There are 3 commands in this PR, 1 in build, 2 in users

mrbelvedereci/build/management/commands/metaci_scheduled_jobs.py
jlantz: done

mrbelvedereci/hirefire/views.py
jlantz: done

Procfile
jlantz: It's added back
https://github.com/SalesforceFoundation/mrbelvedereci/pull/146/commits/f26c16386357ff171fe2c0c220d03b99dbcdaccb

PR #147, opened 2017-10-25
Allow multiple repos per plan

cpolcyn:
Changed Plan.repo (ForeignKey) to Plan.repos (ManyToManyField) with custom through model PlanRepository. Plan add/edit now does not show any repo info. Repos are associated with plans by adding new PlanRepository models.

* new view to list all plans
* manage.py command "compressplans" that merges redundant plans
* `/plans/<id>/run` now asks user to select which repo to run plan against. The full plan run URL is now `/plans/<id>/<repo_owner>/<repo_name>/run/`
* GitHub webhooks are now created on save of PlanRepository using RQ job, instead of in the Plan save handler


mrbelvedereci/plan/management/commands/compressplans.py
```diff
@@ -0,0 +1,96 @@
...
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            default=False,
+            help='Do a dry run (do not modify db)',
+        )
+
+    def handle(self, *args, **options):
+        if options['dry_run']:
+            self.stdout.write('DRY RUN - db will not be modified')
```
cdcarter: can we get a logger in this context?

mrbelvedereci/plan/management/commands/compressplans.py
```diff
@@ -0,0 +1,96 @@
...
+        invalid_plans = set()
+        plan_pks = sorted([plan.pk for plan in Plan.objects.all()])
+        for plan_combo in combinations(plan_pks, 2):
```
cdcarter: consider
`for (plan1pk, plan2pk) in combinations(plan_pks, 2)` ?

mrbelvedereci/plan/management/commands/compressplans.py
```diff
@@ -0,0 +1,96 @@
...
+                'sfdx_config',
+                'yaml_config',
+            ]):
+                continue
+            # keep the plan with more builds
+            if plan1.builds.count() < plan2.builds.count():
+                # swap plans (keep plan1)
+                plan1, plan2 = plan2, plan1
+            # reassign builds to the plan we're keeping (plan1)
+            for build in plan2.builds.all():
```
cdcarter: can we select_for_update (https://docs.djangoproject.com/en/dev/ref/models/querysets/#select-for-update) and run this in a transaction? i'm a little worried about all these build.save() calls taking their own transaction

mrbelvedereci/plan/models.py
```diff
@@ -45,13 +50,23 @@ class Plan(models.Model):
...
     def __unicode__(self):
-        return u'[{}] {}'.format(self.repo.name, self.name)
+        return u'{}'.format(self.name)
+
+    @property
+    def repo_names(self):
+        names = [repo.name for repo in self.repos.all()]
+        return ', '.join(names)
+
+    @property
```
cdcarter: I feel really skeezy having a property return a generator. I would feel more comfortable if it were a method call.

mrbelvedereci/plan/models.py
```diff
@@ -114,6 +129,18 @@ def check_push(self, push):
...
+
+class PlanRepository(models.Model):
+    plan = models.ForeignKey(Plan)
+    repo = models.ForeignKey(Repository)
+
+    class Meta:
+        verbose_name_plural = 'PlanRepositories'
+
+    def __unicode__(self):
+        return u'{}:{}'.format(self.repo, self.plan)
```
cdcarter: changed from `u'[{}] {}'` on purpose?

mrbelvedereci/plan/tasks.py
```diff
@@ -30,3 +33,28 @@ def run_scheduled_daily():
 @django_rq.job('short')
 def run_scheduled_hourly():
     return run_scheduled('hourly')
+
+@django_rq.job('short')
+def create_github_webhook(pk):
+    """Create a webhook if it doesn't exist yet."""
+    plan_repo = PlanRepository.objects.get(pk=pk)
+    if plan_repo.plan.type == 'manual':
```
cdcarter: can this check happen before delaying the job? no need to get a worker all involved...

mrbelvedereci/plan/tasks.py
```diff
@@ -30,3 +33,28 @@ def run_scheduled_daily():
...
+        return
+    event = 'pull_request' if plan_repo.plan.type == 'pr' else 'push'
+    callback_url = '{}/{}'.format(settings.GITHUB_WEBHOOK_BASE_URL, event)
```
cdcarter: urlparse.urljoin

mrbelvedereci/plan/tasks.py
```diff
@@ -30,3 +33,28 @@ def run_scheduled_daily():
...
+    exists = False
+    gh_repo = plan_repo.repo.github_api
+    for hook in gh_repo.iter_hooks():
+        if hook.config.get('url') == callback_url and event in hook.events:
+            exists = True
+            break
+    if not exists:
+        gh_repo.create_hook(
```
cdcarter: i know this isn't new code, so sorry to ask lots of questions in the review.

if this fails, what happens? the worker will throw an exception - but is the sentry context fully initialized here? will i get an email if a webhook fails to create?

cdcarter: this is somewhat important for jason's hypothetical user spinning up a new site.

mrbelvedereci/plan/templates/plan/detail.html
```diff
@@ -92,6 +89,9 @@ <h3 class="slds-heading slds-heading--large">Description</h3>
       <td data-label="Status">
         <div class="slds-truncate slds-badge {% if build.get_status == 'fail' or build.get_status == 'error' %}slds-theme--error{% elif build.get_status == 'success' %}slds-theme--success{% endif %}" title="{{ build.get_status }}">{{ build.get_status }}</div>
       </td>
+      <td data-label="Repository">
+        <div class="slds-truncate" title="{{ build.repo }}">{{ build.repo }}</div>
```
cdcarter: seeing ALL the builds for ALL the repos for a plan seems unhelpful.

i think with this schema, we need to introduce r'^(?P<plan_id>\w+)/(?P<repo_owner>\w+)/(?P<repo_name>\w+)/$'` as a list view to show what plan detail currently does now.

mrbelvedereci/plan/views.py
```diff
@@ -6,10 +6,18 @@
...
+from mrbelvedereci.repository.models import Repository
 
 
-def plan_list(request, plan_id):
-    return ''
+def plan_list(request):
+    if request.user.is_staff:
+        plans = Plan.objects.all()
+    else:
+        plans = Plan.objects.filter(public=True)
```
cdcarter: Confirming what this does for SAL etc....

SAL will still be private at the *REPO* level, but the plan will be public? Logged out users will be able to go to Feature Test plan and see NPSP and HEDA on it, but not SAL?

mrbelvedereci/plan/models.py
cpolcyn: No reason really, just that the other one was going away (plan.__unicode__) so I didn't see a need to keep it the same, but I don't really care so I'll change it back.

mrbelvedereci/plan/tasks.py
cpolcyn: Sure, good call.

mrbelvedereci/plan/templates/plan/detail.html
cpolcyn: I wouldn't say it's unhelpful, it will show the most recent builds of the plan which I'd find useful. But I agree it begs for a way to filter down by repo, and your proposed URL format matches the plan run URL format, I like it. I'll add the repo-specific plan view in addition to this.

mrbelvedereci/plan/views.py
cpolcyn: Good point, but I think in your example wouldn't a logged out user be able to see a SAL plan as long as it was public? I filed this one a while ago and I still think we should implement it: https://github.com/SalesforceFoundation/mrbelvedereci/issues/46 – Even if it doesn't sync to github I still think we should just control public/private at the repo level, but we should continue the debate in that issue. As far as this here is concerned, I think it's consistent with our models, so that if a plan is public but one of the repos it runs against is private, we would show the plan anyway.

cpolcyn: A logged out user will not see any private plans in this view, and public plans with an associated private repo would show that repo in their list, which I think we don't want. So I'll change the repos list to take public/private into account.

mrbelvedereci/plan/management/commands/compressplans.py
cpolcyn: It seems we can get a logger but for some reason it is unable to print. This syntax is recommended in the guide here: https://docs.djangoproject.com/en/1.10/howto/custom-management-commands/

mrbelvedereci/plan/management/commands/compressplans.py
cpolcyn: Actually it's useful to leave this as a tuple for the `set` statement on the next line.

mrbelvedereci/plan/tasks.py
cpolcyn: You're right that it's just kind of lost in a way, at least with it as a rq job it will be put into the failed job queue. Would be good to have Sentry or something send a notification, but I think that's outside the scope of this PR. I've filed https://github.com/SalesforceFoundation/mrbelvedereci/issues/148 to follow up.

mrbelvedereci/plan/templates/plan/detail.html
cpolcyn: Addressed in https://github.com/SalesforceFoundation/mrbelvedereci/pull/147/commits/dbb8fd5014da8be496a4ebff2197f631e1eaec00

mrbelvedereci/plan/views.py
cpolcyn: Addressed in https://github.com/SalesforceFoundation/mrbelvedereci/pull/147/commits/a7aa974593643747d2218b3ff3be4533e88d92a1

mrbelvedereci/plan/management/commands/compressplans.py
cpolcyn: Addressed in https://github.com/SalesforceFoundation/mrbelvedereci/pull/147/commits/3676a6e5686dfd565222c5c25c4e927593cf8e0f

cpolcyn:
@cdcarter I addressed all your comments, please have another look.

metaci/plan/models.py
```diff
@@ -45,14 +51,18 @@ class Plan(models.Model):
...
-        ordering = ['name','repo__owner','repo__name', 'active', 'context']
+        ordering = ['name', 'active', 'context']
 
     def get_absolute_url(self):
         return reverse('plan_detail', kwargs={'plan_id': self.id})
 
     def __unicode__(self):
         return self.name
 
+    def get_repos(self):
```
jlantz: What's the reason for having this vs using the Django standard way of self.repos.all()?  I'm pretty sure self.repos.all() is also a generator.

jlantz: Actually, nevermind... it's to return the repository objects themselves

PR #149, opened 2017-10-26
retry getting org a few times if sfdx fails to create scratch org

### LINKED ISSUE #104: Auto rebuild if Error "Failed to create scratch org"
Automatically rebuild if a build/rebuild is Error due to "Failed to create scratch org". Keep track of rebuilds triggered this way so that we can have a max retries amount. We can reuse "auto" for User like some original builds, but ignore the original builds for counting retries.

cpolcyn:
Since this isn't an exception from the flow we can't catch it in the same way that we do for e.g. Apex test failures. So instead I retry getting the org a few times before giving up, all within the build.

NOTE: depends on https://github.com/SalesforceFoundation/CumulusCI/pull/485

Fixes #104

### DIFF
--- a/mrbelvedereci/build/models.py
+++ b/mrbelvedereci/build/models.py
@@ -274,18 +274,24 @@ class Build(models.Model):
         return project_config
 
     def get_org(self, project_config, retries=3):
-        def get_org_config(attempt=None):
-            if not attempt:
-                attempt = 1
+        logger = init_logger(self)
+        attempt = 1
+        while True:
             try:
-                return project_config.keychain.get_org(self.plan.org)
+                org_config = project_config.keychain.get_org(self.plan.org)
+                break
             except ScratchOrgException as e:
-                if attempt > retries:
+                if (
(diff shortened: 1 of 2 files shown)


mrbelvedereci/build/models.py
```diff
@@ -271,8 +273,19 @@ def get_project_config(self):
         project_config.set_keychain(keychain)
         return project_config
 
-    def get_org(self, project_config):
-        org_config = project_config.keychain.get_org(self.plan.org)
+    def get_org(self, project_config, retries=3):
+        def get_org_config(attempt=None):
```
cdcarter: I think a while True loop would be more easy to quickly understand (like over here https://github.com/SalesforceFoundation/CumulusCI/blob/master/cumulusci/core/tasks.py#L148) but i won't say no to this because of that

cpolcyn:
@cdcarter thanks for the suggestion, I went ahead and made it a while loop. Please have another look as the new commit dismissed your review.

PR #167, opened 2017-11-09
remove creation of webhook for repo

cpolcyn:
Github user needs to be repo admin to create webhook for the repo. This is the only part of the system that requires github repo admin permissions, and is a one-time setup that's easy to do manually.

We want the GIthub user on MetaCI to *not* be admin of repo, because it creates unnecessary risk.

This feature could be re-implemented in MetaCI CLI as the user there may have admin rights on the repo.


cpolcyn:
Will wait on @jlantz to chime in before merging.

jlantz:
Makes sense to me.  I like the idea of doing this in metaci instead of here.  It is important that we make the need to set up the webhook very clear in the CPI and docs though as you don't get the expected functionality with the webhook

PR #171, opened 2017-11-15
Org Market

cdcarter:
Adds a `supertype` column to Org which specifies if its a CI org (MetaCI can run tasks on it) or a Registered Org (an org in the org mart).

Updates the metaci.cumulusci.models.Org entity to add a bunch of attributes (some added to filter):
- org ID (salesforce 15 or 18 digit org id)
- org type (scratch, packaging, persistent unmanaged, persistent beta, persistent prod)
- description
- owner
- release cycle (opt into QA releases for registered orgs only)

Added or updated the following views:
- Updated org list view to have lots more details (and qs filter on type and supertype).
- Added Org ID to display on org details.
- Added `ci_orgs` and `registered_orgs` managers to Org that filter on supertype.
- Current `orgs/` API endpoint switched to only show CI Orgs.
- new `registered_orgs/` API endpoint with CRUD access to register new registered orgs.

The end goal is to easily get a list of orgs that should be used by CCI for QA push upgrade. Well, how's this?
`http://127.0.0.1:8000/api/registered_orgs/?org_type=PRODUCTION&release_cycle=QA&repo=4`


cdcarter:
Okay, changes are wrapped up and this is good to review.

metaci/cumulusci/models.py
```diff
@@ -8,15 +8,54 @@
...
+
+from model_utils.managers import QueryManager
+
+import choices
 
 
 class Org(models.Model):
     name = models.CharField(max_length=255)
     json = models.TextField()
     scratch = models.BooleanField(default=False)
```
cpolcyn: `scratch` is redundant with `org_type`, should be removed.

metaci/cumulusci/admin.py
```diff
@@ -5,8 +5,23 @@
...
 
 class OrgAdmin(admin.ModelAdmin):
-    list_display = ('name', 'repo', 'scratch')
-    list_filter = ('name', 'scratch', 'repo')
+    list_display = ('name', 'repo', 'scratch', 'supertype', 'org_type')
+    list_filter = ('name', 'scratch', 'repo', 'supertype', 'org_type')
+
+    fieldsets = (
+        (
+            None, {'fields':('name','repo','supertype','org_type', 'org_id', 'last_deploy','last_deploy_version')}
```
cpolcyn: Should just show the fields needed to create/edit the object.

cdcarter:
Chatted with Chris offline today and came up with this set of todos:

- [ ] squash migrations
- [ ] validation that CI persistent orgs have persistent org JSON
- [ ] validation that CI scratch orgs have scratch org JSON
- [ ] remove last_release fields
- [ ] add choices to release_cycle (feature commit, master commit, beta release, production release) to categorize CI orgs too (for a self documenting system)
- [ ] review where plan creation and build creation happens so that they can't run against a non-CI org.
- [ ] any other code review notes chris has.

metaci/cumulusci/models.py
```diff
@@ -8,15 +8,54 @@
...
+    last_deploy = models.DateTimeField(null=True, blank=True)
+    last_deploy_version = models.CharField(max_length=255, null=True, blank=True)
+    release_cycle = models.CharField(
+        max_length=50,
+        choices=choices.RELEASE_CHOICES,
+        null=True,
+        blank=True
+    )
+
+    objects = models.Manager()
```
cpolcyn: Remove line?

metaci/cumulusci/models.py
```diff
@@ -50,6 +89,11 @@ def unlock(self):
         if not self.scratch:
             cache.delete(self.lock_id)
 
+    def clean(self):
+        super(Org, self).clean()
```
cpolcyn: Is the `super` line required? Looks like this method is meant to be overridden anyway https://docs.djangoproject.com/en/1.10/ref/models/instances/#django.db.models.Model.clean

cpolcyn:
> add choices to release_cycle

Can't some orgs fall into multiple parts of the cycle? Like the packaging org would be used for beta tags and release.

metaci/cumulusci/models.py
cdcarter: I agree. To avoid a super heavy refactor of every check for it, I'll add a scratch @property to the model

metaci/cumulusci/models.py
cdcarter: you're super right.

cdcarter:
this will be its own repo, someday.

PR #176, opened 2017-11-28
Fix datetime encoding issue with json

jlantz:
In cci beta77+, the OrgConfig.config dict contains a datetime object to track the created date on a scratch org.  That was causing errors like this one when set_org was called on the keychain:
https://mrbelvedereci-staging.herokuapp.com/builds/207


cpolcyn:
Let's use this instead:
https://docs.djangoproject.com/en/dev/topics/serialization/#djangojsonencoder

PR #182, opened 2018-01-04
Fix Scratch Org Login

### LINKED ISSUE #40: Scratch org login not working
Try to log in to an org using the "Log In" button on org tab of build detail view, encounter error 500.

jlantz:
Fixes #40 (I think)

Remove lines that checked for ~/.local/.sfdx and wrote the sfdx org json file there instead of ~/.sfdx  It looks like this was a legacy support issue for older versions of sfdx which is no longer valid and was causing scratch org logins to fail


jlantz:
Tested in a review app and could log into the scratch org!

PR #197, opened 2018-03-02
Feature/qa experience

cdcarter:
Impetus: make it easier for Docs/QA/PM/UX to use manual plans to provision 7 day scratch orgs to test out a branch or latest beta.

* replaces the (empty) bootstrap based user detail view with a user detail view that shows non-deleted scratch orgs owned by the user. it also shows some user details in an SLDS format

screen shot 2018-03-02 at 2 26 25 pm

* creates a new plan trigger type called Org Request to indicate the plan is not only triggered manually, but in cases where the org will stay around for testing.

* adds a "New Org Please" page to allow the easy creation of orgs using Org Request plans by listing all plan repositories

screen shot 2018-03-02 at 2 26 41 pm

* plan run form now accepts query string prefill params. this makes link sharing easier.


cpolcyn:
Since you told me elsewhere that more work is incoming on this, I've removed myself as a reviewer for now. Once it's ready please add me as reviewer again.

metaci/cumulusci/models.py
```diff
@@ -2,6 +2,7 @@
 
 import json
 import os
+import datetime
```
jlantz: I don't see this import being used

metaci/plan/views.py
```diff
@@ -65,10 +65,21 @@ def plan_run_repo(request, plan_id, repo_owner, repo_name):
...
-        form = RunPlanForm(plan, repo, request.user)
+        form = RunPlanForm(plan, repo, request.user, request.GET)
     context = {
         'form': form,
         'plan': plan,
         'repo': repo,
     }
     return render(request, 'plan/run.html', context=context)
+
+@login_required
```
jlantz: This should be restricted to is_staff as anyone can be a logged in user

cdcarter: changed this and the other views for plan_run to match

PR #200, opened 2018-03-09
Add notification type for PlanRepository

cpolcyn:


metaci/notification/forms.py
```diff
@@ -122,6 +123,43 @@ class Meta:
         widgets = {'user': forms.HiddenInput()}
 
 
+class AddPlanRepositoryNotificationForm(forms.ModelForm):
+
+    def __init__(self, *args, **kwargs):
+        super(AddPlanRepositoryNotificationForm, self).__init__(*args, **kwargs)
```
cdcarter: these form helpers are very repetitive for the different types, is there any way we can pull out some of it into a superclass for the different notifications?

metaci/notification/tasks.py
```diff
@@ -36,10 +37,13 @@ def queue_build_notifications(build_id):
     repo_notifications = RepositoryNotification.objects.filter(repo = build.repo, **status_query).values('user__id').distinct()
     branch_notifications = BranchNotification.objects.filter(branch = build.branch, **status_query).values('user__id').distinct()
     plan_notifications = PlanNotification.objects.filter(plan = build.plan, **status_query).values('user__id').distinct()
+    planrepository = PlanRepository.objects.filter(plan=build.plan, repo=build.repo)[0]
```
cdcarter: can we use `PlanRepository.objects.get` instead of the 0 index?

metaci/notification/views.py
```diff
@@ -77,6 +80,23 @@ def add_plan_notification(request):
     )
 
 
+@login_required
```
cdcarter: should notifs be staff_member_required?

metaci/notification/tasks.py
cpolcyn: No because it may return more than 1 object. Although any duplicates would have the exact same info, we currently don't have protection against duplicates existing in the first place.

metaci/notification/forms.py
cpolcyn: In my view the entire notifications system needs an overhaul. To the user there should just be "notifications" and not several different types. This is a significant change, and because this is an urgent fix I opted to follow the established pattern for now. We can do it better if/when we overhaul the notifications.

metaci/notification/views.py
cpolcyn: I don't think so but it's a matter of opinion. In this case I followed the existing pattern. If we want to update it to staff only then I think that's a different PR.

metaci/notification/tasks.py
cdcarter: :( fair enough. we should definitely have a unique index there, but not in scope this PR.

metaci/notification/views.py
cdcarter: fine by me, but we really need to take a look at our permissions. i'm worried if we can just send email to anyone.

PR #202, opened 2018-03-13
fix template render issue in notifications

cpolcyn:
Can no longer pass Context object to Template, should be dict. Simplified by using [render_to_string](https://docs.djangoproject.com/en/1.11/topics/templates/#django.template.loader.render_to_string).


cpolcyn:
https://docs.djangoproject.com/en/1.11/releases/1.10/#features-removed-in-1-10

cpolcyn:
We were on Django 1.10 until recently so I'm surprised this issue has only popped up now that we're on 1.11.

PR #211, opened 2018-03-14
Feature/jwt scratchorg login

cdcarter:
the JSON DX is now encrypted and we are unable to simply write it to disk and use the DX client. This PR stops writing json_dx to the database and doesn't write to disk either. instead, we establish a JWT grant with the target org to login or delete.

*eventually, the JWT code must be refactored into the keychain, but likely in CCI core instead of in metaci.* i'd like to chat w/ lantz about this, because my grip of what belongs where is tenuous.

so, this PR changes ScratchOrgInstance.delete_org (used in all scratch org deletion) to JWT to dev hub and delete the org. It also changes the scratch org login view to generate a JWT session and knock on the frontdoor.

this now requires three additional environment variables be set/added to your local_setup.sh. These env vars are already set in staging and prod (client id, hub key, hub username)


metaci/cumulusci/models.py
```diff
@@ -11,6 +16,46 @@
...
+    response.raise_for_status()
+
+    return response.json()
+
+def sf_session(jwt):
+    return SimpleSalesforce(
+            instance_url=jwt['instance_url'],
+            session_id=jwt['access_token'],
+            client_id='metaci',
+            version='42.0'
```
cpolcyn: We should probably use an env var here but I don't see one already so I won't hold up the PR for that.

metaci/cumulusci/models.py
```diff
@@ -75,29 +119,38 @@ def get_absolute_url(self):
...
-            f.write(self.json_dx)
+        return self._get_org_config()
 
+    def _get_org_config(self):
         org_config = json.loads(self.json)
         org_config['date_created'] = parse_datetime(org_config['date_created'])
         return ScratchOrgConfig(org_config, self.org.name)
 
+    def get_jwt_based_session(self):
+        return jwt_session('https://test.salesforce.com', self.username)
```
cpolcyn: Might be clearer to use a sandbox flag instead of passing the whole URL?

metaci/cumulusci/models.py
cdcarter: for sure.

metaci/cumulusci/models.py
cdcarter: yes. yes it would be :)

PR #214, opened 2018-03-15
add rq job to mark branches as deleted

cpolcyn:
Overrides Branch.objects to use a custom manager with filter(deleted=False)

Once this is deployed, we can set up a repeatable job that calls this to periodically prune the deleted branches.


cpolcyn:
Note that the original Branch.objects manager is still available as Branch._base_manager

https://docs.djangoproject.com/en/1.11/topics/db/managers/#django.db.models.Model._base_manager

cdcarter:
looks good. 

we could rename the field (but not the db_column) to `is_removed` and then use `SoftDeletableManager`(https://github.com/jazzband/django-model-utils/blob/master/model_utils/managers.py#L327) from django_model_utils, which allows 
```
branch.delete() 
```
to work directly. Not sure if it's worth it...

or, you could collect the ids to update and then do a bulk update. this is just gonna thrash the db the first time its runs I guess...

cpolcyn:
Thanks for the reminder about django-model-utils. That's exactly what I need. Instead of dealing with the manager directly, I just updated Branch to inherit from SoftDeletableModel, and updated the task to use normal delete() method.

cpolcyn:
@cdcarter my new commits dismissed your review. Can you please re-review?

PR #223, opened 2018-03-20
Added additional default plans to the migration that creates them.

jlantz:
Since this modifies an existing migration and the logic doesn't run if
any plans already exist, this shouldn't have any impact on existing
sites.

New Plans:
* Upload Release
* Release Test
* Dev Org

NOTE: The inclusion of `setuptools` in the requirements/base.txt file was done to fix a build error on Heroku with installing `django-watson`.  Per the following doc, setuptools is missing in the build environment causing the error.  Adding it allowed the build to pass.
https://stackoverflow.com/questions/11425106/python-pip-install-fails-invalid-command-egg-info#11425830

Failing test for this branch before the change:
https://dashboard.heroku.com/pipelines/65857394-4132-4e78-aa40-ad30b09ae8da/tests/372

Failing test of master showing the same error which this PR should also fix in master:
https://dashboard.heroku.com/pipelines/65857394-4132-4e78-aa40-ad30b09ae8da/tests/374


metaci/plan/migrations/0013_create_default_plans.py
```diff
@@ -49,6 +49,43 @@ def create_default_plans(apps, schema_editor):
...
+        context='Upload Release',
+    )
+    # Release Test
+    Plan.objects.create(
+        name='Release Test',
+        description=(
+            'Installs the latest release in the standard release test org and '
+            'runs apex tests'
+        ),
+        type='tag',
```
cdcarter: should release test set a github context as well?

PR #231, opened 2018-04-11
 Scratch Org Deletion & ui...

cdcarter:
This is a bit of a mishmash, happy to break it into two PRs but i just wanted to get it out of my working dir while I'm working on robot import stuff.

The main bulk of the PR
- Adds an `expiration_date` column to ScratchOrgInstance that is populated on org create.
- Adds new managers to filter on expiration date of SOI.
- Adds a pruning task (based on Chris's branch pruning task) to move expired orgs into deleted.
- Updates user profile page to use new manager.

Additionally...:
- Adds the plan description to the plan run page, making it easier to see what you're about to do.
- marks `build` as a raw ID field in the django admin on Scratch Org Instance, which makes it...even possible to load an SOI without timeouts.
- Added a "Orgs" tab to the repo detail page (which...could use a refactor). The tab just has the org name and a link to it, but other ways of getting to an org detail page are wacky, so I find it useful enough.


metaci/cumulusci/tasks.py
```diff
@@ -0,0 +1,19 @@
...
+
+from metaci.cumulusci.models import ScratchOrgInstance
+
+
+@job('short')
+def prune_orgs():
+    db.connection.close()
+    pruned = []
+    for soi in ScratchOrgInstance.expired.all():
+        soi.delete()
```
cpolcyn: This actually deletes the row since the model is a standard Django model. Did you mean to use `soi.delete_org()` instead?

cdcarter: whoops good catch yes, originally I had looked at switching SOI to a softdeletemodel but the mechanics of scratch orgs are a little weird, so it didn't make sense in the end.

cdcarter:
ready for rereview @cpolcyn

metaci/repository/models.py
```diff
@@ -42,7 +42,7 @@ def get_absolute_url(self):
         return reverse('branch_detail', kwargs={'owner': self.repo.owner, 'name': self.repo.name, 'branch': self.name})
 
     def __unicode__(self):
-        return u'[{}] {}'.format(self.repo.name, self.name)
+        return u'{}'.format(self.name)
```
cpolcyn: This was good to have for selecting a branch in a dropdown menu, e.g. when running a plan. Can we change this back?

cpolcyn: Just saw this isn't used in plan run view anyway, so it seems safe to change.

PR #238, opened 2018-05-10
Require cumulusci-2.0.0b94

cpolcyn:


cdcarter:
Awesome @cpolcyn - lets get this into MetaCI staging and do an end to end of CumulusCI-Test just to make sure that nothing got weird in our normal deployment flows...

PR #243, opened 2018-05-21
Task Results in Database

cdcarter:
This PR takes advantage of BaseFlow hooks introduced around CCI2b94.

if `METACI_FLOW_SUBCLASS_ENABLED` is True, the app will use MetaCIFlow instead of BaseFlow as its flow runner. the flow will update FlowTask objects for each executing task.

this data is collected in the database but not displayed on the UI (yet). this opens up a few things:
- using task return values/results for more than just passing params around
- reporting on task runtime 
- visualizing a build


config/settings/local.py
```diff
@@ -103,3 +103,4 @@
 SFDX_HUB_USERNAME = env('SFDX_HUB_USERNAME')
 
 GITHUB_STATUS_UPDATES_ENABLED = env.bool('GITHUB_STATUS_UPDATES_ENABLED', False)
+METACI_FLOW_SUBCLASS_ENABLED = env.bool('METACI_FLOW_SUBCLASS_ENABLED', True)
```
cpolcyn: Can we remove this line and just leave default=False for all configs?

config/settings/local.py
```diff
@@ -103,3 +103,4 @@
 SFDX_HUB_USERNAME = env('SFDX_HUB_USERNAME')
 
 GITHUB_STATUS_UPDATES_ENABLED = env.bool('GITHUB_STATUS_UPDATES_ENABLED', False)
+METACI_FLOW_SUBCLASS_ENABLED = env.bool('METACI_FLOW_SUBCLASS_ENABLED', False)
```
cpolcyn: Might at well remove this since it's same as in common.py

PR #250, opened 2018-06-01
do not rely on pip for cci version

cdcarter:


cdcarter:
well that worked on my local... lol. i'll try and figure it out.

PR #264, opened 2018-07-26
Fix junit parsing to work with robot junit output

jlantz:


metaci/build/models.py
```diff
@@ -528,11 +528,12 @@ def load_junit(self, filename):
                 if element.tag not in ['failure', 'error']:
                     continue
                 result['Outcome'] = 'Fail'
-                result['StackTrace'] += element.text + '\n'
+                if element.text:
+                    result['StackTrace'] += element.text + '\n'
                 message = element.attrib['type']
-                if 'message' in element.attrib:
+                if 'message' in element.attrib and element.attrib['message']:
```
davisagli: I think you can just do `if element.get('message')`

jlantz: wouldn't it have to be element.attrib.get('message') or does ET provide a get method for attributes?

davisagli: https://docs.python.org/2/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.get

PR #265, opened 2018-07-27
Require cumulusci 2.0.3

jlantz:


runtime.txt
```diff
@@ -1 +1 @@
-python-2.7.14
+python-2.7.15
```
davisagli: Should also update the installation docs

davisagli: Wait no, that's for CCI itself, never mind

jlantz: It's the version for Heroku to run in its dyno :)

PR #275, opened 2018-08-09
Pass the right build id to set_github_status

davisagli:


metaci/build/models.py
```diff
@@ -410,7 +410,7 @@ def run(self, project_config, org_config):
 
         # Update github status
         if settings.GITHUB_STATUS_UPDATES_ENABLED:
-            set_github_status.delay(self.id)
+            set_github_status.delay(self.build.id)
```
cdcarter: use `self.build_id` to avoid a join.

davisagli: Thanks, I forgot that Django adds that attribute automagically.

PR #276, opened 2018-08-15
Feature/releaseapp

cdcarter:
opening this for a PR app... Comments are appreciated, and I'll request a review when I have finished initial testing.

Add a new app to MetaCI to track Releases! This is going to start out lightweight, and grow with time as we automate the process. 

The impetus behind this initial PR is to have a Release record in MetaCI that we can relate Release Test Results to for a test dashboard. When a push event is received for a production release tag (regex defined on Repo), the webhook will auto-create a Release record withe the details we can glean from the tag.

Instead of JUST being a repo and a name at this time, I've added the basic fields that we fill out on the Release Checklist which release engineering can fill out in the django admin instead. providing a relational store for this data potentially provides a better backend for statusite than the plain github releases, for example. 

This is not yet a prescriptive design for release automation in MetaCI. stay tuned...


cdcarter:
:bug: "Builds that are triggered from a push event that matches the release regex will be automatically connected to the release." doesn't work quite yet...

metaci/build/migrations/0015_auto_20180815_2300.py
```diff
@@ -0,0 +1,27 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='build',
+            name='release',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to='release.Release'),
+        ),
+        migrations.AddField(
+            model_name='build',
+            name='release_relationship_type',
+            field=models.CharField(blank=True, choices=[('test', 'Release Test'), ('automation', 'Release Automation')], max_length=50, null=True),
```
davisagli: This is missing the Manual option that's included in models.py

metaci/plan/forms.py
```diff
@@ -16,37 +16,52 @@ class RunPlanForm(forms.Form):
...
         super(RunPlanForm, self).__init__(*args, **kwargs)
         self.fields['branch'].choices = self._get_branch_choices()
+        self.fields['release'].queryset = self.repo.releases
         self.helper = FormHelper()
         self.helper.form_class = 'form-vertical'
         self.helper.form_id = 'run-build-form'
         self.helper.form_method = 'post'
+        self.advanced_mode = False
+        if 'advanced_mode' in args[0]:
+            self.advanced_mode = (args[0]['advanced_mode'] == "1")
```
davisagli: Can we include a link to switch to advanced mode somewhere on the page? I think it's fine to have an advanced mode triggered by a URL param, but I'd rather it not be hidden from anyone who wasn't told or has forgotten it exists.

metaci/build/models.py
```diff
@@ -119,6 +124,9 @@ class Build(models.Model):
     build_type = models.CharField(max_length=16, choices=BUILD_TYPES, default='legacy')
     user = models.ForeignKey('users.User', related_name='builds', null=True, on_delete=models.PROTECT)
 
+    release_relationship_type = models.CharField(max_length=50, choices=RELEASE_REL_TYPES, null=True, blank=True)
```
davisagli: I get the distinction between manual and automated, but what's the distinction between Release Test and Release Automation about?

metaci/release/models.py
```diff
@@ -0,0 +1,40 @@
...
+class Release(StatusModel):
+    STATUS = Choices('draft', 'published', 'hidden')
+    created = AutoCreatedField(_('created'))
+    modified = AutoLastModifiedField(_('modified'))
+
+    repo = models.ForeignKey('repository.Repository', 
+                            on_delete=models.CASCADE, related_name="releases")
+    
+    version_name = models.CharField(_('version name'), max_length=255, null=True, blank=True)
+    version_number = models.CharField(_('version number'), max_length=255, null=True, blank=True)
```
davisagli: What's the version name?

metaci/repository/views.py
```diff
@@ -199,6 +200,21 @@ def github_push_webhook(request):
             branch.is_removed = False
             branch.save()
 
+    release = None
+    # Check if the event was triggered by a tag
+    if push['ref'].startswith('refs/tags/') and repo.release_tag_regex:
+        tag = push['ref'][10:]
```
davisagli: Maybe write `[len('refs/tags/'):]` so it's less of a magic number?

metaci/repository/views.py
```diff
@@ -199,6 +200,21 @@ def github_push_webhook(request):
...
+        # Check the tag against regex
+        if re.match(repo.release_tag_regex, tag) and push['head_commit']:
+            release, _ = Release.objects.get_or_create(
+                repo=repo,
+                git_tag=tag,
+                defaults = {
+                    'created_from_commit': push["head_commit"]["id"],
+                    'status': 'draft'
+                }
+            )
```
davisagli: It looks like if a tag is created and we create and publish a Release, and then the tag is force-changed to a different commit, that would trigger resetting the existing Release to draft status. I'm not sure we need to handle this case but maybe it would be better to abort if a release already exists for the tag?

metaci/release/models.py
cdcarter: They're technically two different things on a Salesforce package, though we have generally kept the our version names to be Product Name + Version Number. This could probably be an admin pre-fill feature

metaci/build/migrations/0015_auto_20180815_2300.py
cdcarter: Good catch. makemigrations didn't care enough to build a new migration for it, but I'll include it manually.

metaci/build/models.py
cdcarter: Release Test = a build related to the output of a release - runs all tests on managed package for example

Release Automation = a build related to the creation of a release, once we start automating those

metaci/repository/views.py
cdcarter: Oh yes, this is weird.

cdcarter: Upon further examination and testing - this won't reset the existing Release in any way, but the new tag information will be thrown away.

cdcarter:
Ready rereview @davisagli

metaci/repository/views.py
davisagli: Okay, makes sense.

Issue #268, opened 2018-07-31
Store robot log.html and screenshots


jlantz:
The log.html that Robot Framework creates from tests executions is a very useful view of the details of test execution.  The execution can also capture screenshots.  None of this is currently captured in a build.

This will likely need to be stored on S3 via django-storages and boto which isn't currently integrated with MetaCI.

Also, we need to be careful that we don't expose anything sensitive in the log files such as access tokens to orgs which currently show up in the log.html.

jlantz:
Implemented

PR #283, opened 2018-08-23
Parse and store robotframework's output.xml file in TestResults

jlantz:
This branch integrates robotframework test results with MetaCI's TestResults model as part of the implementation for #268 :

- If an output.xml file is found in the build directory, parse it into individual xml strings for each test case and store the results in the new `robot_xml` field on TestResult
- Add a new `test_type` field on TestClass
- Add two new views: `test_result_detail` and `test_result_robot`

Additionally, this branch expands the enforcement of build visibility via the `build.utils.view_queryset` filtered queryset for builds based on user access.  This allows opening up the test trend view which now filters out results from build the user shouldn't be able to see.


davisagli:
How can I see this in action? https://mrbelvedereci-staging-pr-283.herokuapp.com/builds/ shows nothing and I get an error when I try to log in.

jlantz:
Unfortunately there's not currently a good way to do that short of deploying it to staging.  If the code looks correct, we can merge it to master to deploy to staging where we can see real builds.  I'm really wanting to get this on our production site today if possible so I'll handle any issues on staging ASAP.

metaci/testresults/importer.py
```diff
@@ -56,20 +60,18 @@ def import_test_results(build_flow, results):
 
         testclass = classes.get(result['ClassName'], None)
         if not testclass:
-            try:
-                testclass = TestClass.objects.get(name = result['ClassName'], repo=build_flow.build.repo)
-            except TestClass.DoesNotExist:
-                testclass = TestClass(name = result['ClassName'], repo=build_flow.build.repo)
-                testclass.save()
+            testclass, created = TestClass.objects.get_or_create(
+                name = result['suite']['name'],
```
davisagli: I don't think this exists for non-robot tests imported using load_junit

metaci/testresults/importer.py
```diff
@@ -56,20 +60,18 @@ def import_test_results(build_flow, results):
...
 
         method = methods.get(class_and_method, None)
         if not method:
-            try:
-                method = TestMethod.objects.get(testclass = testclass.id, name = result['Method'])
-            except TestMethod.DoesNotExist:
-                method = TestMethod(testclass = testclass, name = result['Method'])
-                method.save()
+            method, created = TestMethod.objects.get_or_create(
+                testclass = testclass, name = result['name']
```
davisagli: Ditto

metaci/testresults/importer.py
```diff
@@ -121,3 +123,102 @@ def populate_limit_fields(testresult, code_unit):
...
+            outcome = result['status'],
+            source_file = result['suite']['file'],
+            robot_xml = result['xml'],
+        )
+        testresult.save()
+
+def parse_robot_output(path):
+    """ Parses a robotframework output.xml file into individual test xml files """
+    
+    tree = ET.parse(path)
```
davisagli: Not a big deal here since the xml is machine-generated, but let's get in the habit of using cumulusci.utils.elementtree_parse_file so that the exception message includes the filename when there's a parse error

metaci/testresults/utils.py
```diff
@@ -10,6 +10,8 @@ def find_buildflow(request, build_id, flow):
 
     if not build.plan.public and not request.user.is_staff:
         raise PermissionDenied()
+    if not build.repo.public and not request.user.is_staff:
+        raise PermissionDenied()
```
davisagli: Is this supposed to be different from the previous conditional?

davisagli: Never mind, it is.

metaci/testresults/importer.py
jlantz: I added a new JUnit test_type choice and changed import_test_results to require a test_type value.  Tests loaded via JUnit will get the test_type JUnit now.

metaci/testresults/importer.py
jlantz: I'm not sure what's the issue here.  TestMethod doesn't have a test_type field on it.  The only thing I changed was to use get_or_create instead of try/except DoesNotExist.

metaci/testresults/importer.py
davisagli: My point is that `load_junit` returns a dict that does not contains a `suite` key.

metaci/testresults/importer.py
jlantz: Done.  I also swapped out usage of ET.parse in BuildFlow.load_junit

metaci/testresults/importer.py
jlantz: Ah, I see.  Fixed in latest commit.

PR #284, opened 2018-08-24
Handle robot test output files in the app root directory

jlantz:
Not certain why this is necessary yet, but it does work on my local instance and should fix the same issue in staging

### DIFF
--- a/metaci/build/models.py
+++ b/metaci/build/models.py
@@ -230,7 +230,7 @@ class Build(models.Model):
                     flow=flow,
                 )
                 build_flow.save()
-                build_flow.run(project_config, org_config, self)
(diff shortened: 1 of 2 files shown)


metaci/build/models.py
```diff
@@ -194,6 +194,7 @@ def run(self):
         try:
             # Extract the repo to a temp build dir
             self.build_dir = self.checkout()
+            self.root_dir = os.getcwd()
```
davisagli: I think it'd be better to avoid storing root_dir on the build instance (since it's not a db field) and just pass root_dir to run() instead of the entire build instance.

jlantz: fixed in latest commit

PR #286, opened 2018-08-24
Store robot screenshots on Amazon S3 in production

jlantz:
and rewrite robot's output.xml with correct urls


metaci/testresults/importer.py
```diff
@@ -158,6 +163,26 @@ def import_robot_test_results(build_flow, path):
         )
         testresult.save()
 
+        if result['screenshots']:
+            dirname = os.path.dirname(path)
+            for screenshot in result['screenshots']:
+                screenshot_path = screenshot
+                if dirname:
```
davisagli: When would this not be true?

metaci/testresults/importer.py
```diff
@@ -193,10 +218,18 @@ def get_robot_tests(root, elem, parents=[]):
...
                 'status': 'Pass' if status.attrib['status'] == 'PASS' else 'Fail',
+                'screenshots': [],
             }
             start = datetime.strptime(status.attrib['starttime'], '%Y%m%d %H:%M:%S.%f')
             end = datetime.strptime(status.attrib['endtime'], '%Y%m%d %H:%M:%S.%f')
             delta = end - start
+           
+            # Process screenshots
+            for msg in test.findall(".//msg[@html='yes']"):
+                txt=''.join([text for text in msg.itertext()])
```
davisagli: You can just do `txt = ''.join(msg.itertext())` according to elementtree docs

metaci/testresults/importer.py
jlantz: If the robot output files are located in the build dir rather than in app root, the dirname would be empty.  In that case we just pass the name of the file.  The dirname is only passed if the files are found in the app root dir, otherwise `screenshot` is just the file name with no path before it.  The need for this would go away if we can figure out why robot started writing these files into the app root instead of the temp build dir.

davisagli: Got it, dirname ends up empty if `path` is just a filename with no directory--that's the case I was overlooking. I did look a bit to see why robot is writing to the app root, but it's not immediately obvious.

PR #288, opened 2018-09-07
QA Workflow

### LINKED ISSUE #244: Repo Column Is Showing Different Data on New Org Page vs Profile Page (column titles may be reversed on New Org page)
Currently, the "New Org" and "Profile" pages show different data in the "Repo" columns.  (It appears that the title for the "Repo" and "Plan" columns may be reversed on the "New Org" page.)

metaci - profile page

metaci - new org please page

jlantz:
This branch extends the functionality to manually create a scratch org to support a QA workflow which runs the build, keeps the org, and provides a form for QA to set status, qa comment, and delete the org.

- Add new Build.status choice `qa`
- Add new Plan.type choice `qa`
- Add the QA tab to builds where plan is of type `qa` and when the initial build has completed

To support the UX, the New Org, Please functionality has been renamed and reworked a bit
- New top level nav item Create Org if user is_staff
- New Create Org page with links to Create Scratch Org and Create QA Org
- Removed New Org, Please button

Also fixes #244


jlantz:
Some screenshots from local dev instance...

Create Org:
image

Build in QA state:
image

QA Tab:
image

QA Tab after submitting:
image

config/urls.py
```diff
@@ -31,6 +31,7 @@
 
     url(r'^api/', include('metaci.api.urls')),
     url(r'^builds/', include('metaci.build.urls')),
+    url(r'^create-org/', include('metaci.create_org.urls')),
```
cdcarter: sentimentally, I might ask that we keep new-org-plz as the URL path or offer it was an easter egg.

metaci/build/forms.py
```diff
@@ -0,0 +1,64 @@
...
+            Field('delete_org', css_class='slds-checkbox'),
+        )
+        self.helper.layout.append(
+            FormActions(
+                Submit('submit', 'Submit',
+                       css_class='slds-button slds-button--brand')
+            )
+        )
+
+    def save(self):
```
cdcarter: should this persist to a TestResult instead of properties on the Build?

metaci/build/migrations/0020_auto_20180907_1414.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.10 on 2018-09-07 14:14
+from __future__ import unicode_literals
```
cdcarter: ideally squash the migrations before we push to prod?

metaci/build/models.py
```diff
@@ -109,6 +110,8 @@ class Build(models.Model):
     log = models.TextField(null=True, blank=True)
     exception = models.TextField(null=True, blank=True)
     error_message = models.TextField(null=True, blank=True)
+    qa_comment = models.TextField(null=True, blank=True)
+    qa_user = models.ForeignKey('users.User', related_name='builds_qa', null=True, on_delete=models.PROTECT)
```
cdcarter: related name isn't very fluent to read, qa_builds seems more logical...

metaci/plan/views.py
```diff
@@ -75,12 +75,23 @@ def plan_run_repo(request, plan_id, repo_owner, repo_name):
...
     context = {
         'plans': plans,
         'plan_repos': plan_repos,
     }
-    return render(request, 'plan/new_org_please.html', context=context)
\ No newline at end of file
+    return render(request, 'plan/scratch_org.html', context=context)
+
+@staff_member_required
+def qa_testing(request):
```
cdcarter: whats the difference between this method and the one in metaci/create_org/views.py?

metaci/plan/forms.py
```diff
@@ -32,20 +32,22 @@ def __init__(self, plan, repo, user, *args, **kwargs):
         self.advanced_mode = False
         if 'advanced_mode' in args[0]:
             self.advanced_mode = (args[0]['advanced_mode'] == "1")
+        else:
+            self.fields['keep_org'].widget = forms.HiddenInput()
```
cdcarter: move this down to point where we actually branch on `if self.advanced_mode` to keep the behavior seperate from where we detect it

metaci/build/forms.py
```diff
@@ -0,0 +1,64 @@
...
+        result = self.cleaned_data.get('result')
+        comment = self.cleaned_data.get('comment')
+        delete_org = self.cleaned_data.get('delete_org')
+        if self.build.current_rebuild:
+            build = self.build.current_rebuildQ
```
cdcarter: Q ?

metaci/build/models.py
jlantz: I personally like this format better though.  If you're looking through an alpha ordered list of attrs, you'd see builds and builds_qa next to each other which seems easier to work with.

metaci/build/forms.py
jlantz: TestResult is intended for automated tests and requires a method and class which seems like a different use case.

config/urls.py
jlantz: I get the desire for sentimentality but value UX consistency (including url paths) more.  After all, we did drop the most sentimental thing from MetaCI altogether, it's original name :)

metaci/build/models.py
cdcarter: my only comment is musing on better django tooling to describe models the way desrcibe sobject works so that you see child relationships all together. BUT. that's not what we're here to discuss. fine by me. carry on.

metaci/build/forms.py
jlantz: Typo, resolved in latest commit.  I didn't do a local test with a rebuild since this was introduced apparently.  Just tested locally and it worked though I found one minor bug and fixed it in the QA tab template.

metaci/plan/views.py
jlantz: None, deleted in latest commit

metaci/plan/forms.py
jlantz: Done in latest commit

metaci/build/migrations/0020_auto_20180907_1414.py
jlantz: Squashed migration is in the package.  I never even knew about squashmigrations :)

PR #292, opened 2018-09-14
Feature/planrepoadmin

cdcarter:
- Add PlanRepository.active boolean field (defaults to True, since we've been deleting inactive planrepos) to determine if a planrepository should be ...active
- Only evaluate webhooks for active planrepos
- Only display active planrepos for the run plan screen (and new org please)
- Display the active status on planrepo list views (on the plan and repo details pages)
- Add a PlanRepositoryInline to both the plan and repository admin configs, making it less tedious to wire up a new repo.
- remove the deprecated PlanSchedule model and table, lest we try and do something like add a row to that model and it crash and burn


metaci/repository/templates/repository/repo_plans.html
```diff
@@ -69,15 +76,15 @@
             <div class="slds-truncate"><span>{{ plan.context }}</span></div>
           </th>
           <th data-label="Active">
-            <div class="slds-truncate"><span>{{ plan.active }}</span></div>
+            <div class="slds-truncate"><span>{{ planrepo.active }}</span></div>
```
davisagli: What if the planrepo is active but the plan is not? Then it'll look active here even though the github hook will skip it.

cdcarter: good catch.

cdcarter:
okay, I don't actually like this because you can't .filter() on `alive`, you have to know the private member. so...not totally done.

cdcarter:
@davisagli - okay, the default manager for PlanRepository now automatically adds a calculated `alive` based on both plan and planrepo.  I display this on the Repository plan list. I also use this when filtering on plan repo now.

buuut...it doesn't work on the related filter in the webhook, so...not quite there yet.

even then, I did something hackish to make the pointwise version "work", see the property on the model. I'm...not sure this is a good idea? there are definitely going to be a bunch of ways that it could get stale...

going to take a bit more iteration... probably should have tests, probably should have tests around everything around it.... but bare minimum i need to fix the related filter on the webhook.

metaci/plan/models.py
```diff
@@ -121,19 +121,27 @@ def check_push(self, push):
...
-)
 
+class PlanRepositoryManager(models.Manager):
+    def get_queryset(self):
+        return super(PlanRepositoryManager, self).get_queryset().annotate(
+            alive=models.ExpressionWrapper(
+                models.Q(active=True) & models.Q(plan__active=True),
+                output_field=models.BooleanField()
+            )
+        )
```
davisagli: I think a comment explaining what's going on here would be helpful to our future selves.

davisagli:
Note that the migration failed in HerokuCI

cdcarter:
@davisagli - got rid of that wacky computed field, and used a queryset method as discussed. also resolved the CI issue, and added some more validations that the planrepo is indeed active.

cdcarter:
ready for rereview :)

PR #297, opened 2018-09-20
Fix build success status description

jstvz:


davisagli:
Thanks @jstvz! I just hit the button to deploy since no builds are running at the moment.

PR #306, opened 2018-10-19
Use django-guardian for all front end permissioning

jlantz:
Change all permissions to use django-guardian object level permissions on the front end.  Most everything (querysets and views) are now secured by the following permissions on `PlanRepository`:

* run_plan
* view_builds
* rebuild_builds
* qa_builds
* org_login

The migration scripts will create groups with global access to those permissions and convert all existing `is_staff` users to members of those groups to replicate the current functionality.  More granular control can then be done by changing group assignments for users.  You can use per PlanRepository permissions by creating Groups and granting the individual permissions to groups by going to the PlanRepository in admin and using the new Object Permissions button.

The migration will also grant AnonymousUser the view_builds perm on all PlanRepository objects on a public Plan of a public Repository to retain the current anonymous user access to builds.

Removed the fields Plan.public and Repository.public as they are no longer needed for permission enforcement.

# WARNING re: Org.management_groups
This PR removes the Org.management_groups field and it's prior functionality.  The same functionality can be accomplished via object level permissions but it wasn't possible to figure out a clean migration path for that.  If a user has the global `org_login` permission, they can log into all orgs.  The recommended path is to not grant `org_login` globally to users and to instead grant it through object level permissions assigned to groups.


jlantz:
I need to do a bit of work on the build/detail_layout.html view which is not showing the right tabs and buttons but think this is pretty much good to review beyond that.

davisagli:
Can we rename `view_builds_org` to something like `org_login`? "view" makes it sound more innocuous than it is.

metaci/cumulusci/views.py
```diff
@@ -17,14 +16,22 @@
...
 
 
-@staff_member_required
+@login_required
 def org_detail(request, org_id):
-    org = get_object_or_404(Org, id=org_id)
+    try:
+        org = Org.objects.for_user(request.user).get(id=org_id)
+    except Org.DoesNotExist:
+        raise Http404
```
davisagli: Maybe we'll want a `get_obect_for_user_or_404` helper?

metaci/templates/layout_full.html
```diff
@@ -87,7 +87,7 @@ <h3 class="slds-text-heading--medium">Create Org</h3>
         </div>
         </a>
       </li>
-      {% if request.user.is_superuser %}
+      {% if request.user.is_staff %}
```
davisagli: Not still superuser?

jlantz: No because is_superuser isn't really right for how Django's admin access works.  The purpose of is_staff is to control if you have access to admin.  If you only have is_staff and no perms, you can get to admin but not see anything.  In our migration scripts, we remove is_staff from users after adding them to groups that grant what is_staff used to grant on the front end.  This config allows you to grant users is_staff and, for example, the ability to edit only Orgs.

metaci/cumulusci/views.py
jlantz: I was thinking about that too

metaci/build/models.py
```diff
@@ -132,8 +142,13 @@ class Build(models.Model):
     release_relationship_type = models.CharField(max_length=50, choices=RELEASE_REL_TYPES, null=True, blank=True)
     release = models.ForeignKey('release.Release', on_delete=models.SET_NULL, null=True, blank=True)
 
+    objects = BuildQuerySet.as_manager()
+
     class Meta:
         ordering = ['-time_queue']
+        permissions = (
+            ('search_builds', 'Search Builds'),
```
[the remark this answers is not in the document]
jlantz: Yes, I'm going to assign it in the View All Builds group since I don't think we do perm filtering on search at the moment so you really need the global view_builds perm to be able to use search.

metaci/templates/layout_full.html
davisagli: Ah, yep.

metaci/build/models.py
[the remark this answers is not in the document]
jlantz: commit pushed

metaci/build/views.py
```diff
@@ -29,21 +30,13 @@ def build_list(request):
...
-                'You are not authorized to view this build')
-        if tab == 'org':
-            return HttpResponseForbidden(
-                "You are not authorized to view this build's org info")
-        if tab == 'qa':
-            return HttpResponseForbidden(
-                "You are not authorized to view this build's QA info")
+    if not request.user.has_perm('plan.view_builds', build.planrepo):
+        return HttpResponseForbidden(
+            'You are not authorized to view this build')
```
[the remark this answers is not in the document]
jlantz: Well, the view doesn't show links to views you don't have access to.  So the only way to get to these views would be either url hacking or potentially if someone shared the url to the view with them via copy/paste

metaci/build/views.py
```diff
@@ -77,51 +70,88 @@ def build_detail(request, build_id, rebuild_id=None, tab=None):
...
+    else:
+        form = QATestingForm(build, request.user)
+    context['form'] = form
+    return render(request, 'build/detail_qa.html', context=context)
 
 def build_rebuild(request, build_id):
     build = get_object_or_404(Build, id=build_id)
 
-    if not request.user.is_staff:
+    if not request.user.has_perm('plan.qa_builds', build.planrepo):
```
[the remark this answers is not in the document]
jlantz: commit pushed

metaci/cumulusci/models.py
```diff
@@ -55,20 +57,25 @@ def sf_session(jwt):
...
+    def for_user(self, user, perms=None):
+        if perms is None:
+            perms = 'plan.view_builds_org'
+        PlanRepository = apps.get_model('plan.PlanRepository')
+        planrepos = PlanRepository.objects.for_user(user, perms)
+        planrepos = planrepos.values('plan__org', 'repo')
+        q = models.Q()
+        for plan_org in planrepos:
+            q.add(models.Q(name=plan_org['plan__org'], repo_id=plan_org['repo']), models.Q.OR)
+        return self.filter(q)
```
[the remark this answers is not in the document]
jlantz: We could build a view for that, but this is at least better than it was before.

metaci/cumulusci/views.py
jlantz: Added in latest commit

davisagli:
@jlantz yeah I agree we're going to want the granularity, I just suspect we'll need a better way to manage it.

metaci/cumulusci/views.py
```diff
@@ -17,14 +16,21 @@
 from metaci.cumulusci.models import Org
 from metaci.cumulusci.models import ScratchOrgInstance
 from metaci.cumulusci.utils import get_connected_app
+from metaci.plan.models import PlanRepository
 
 
-@staff_member_required
+@login_required
```
cdcarter: i'm pretty sure login_required sends people to the admin login screen, not through social/our templates. this is going to be more annoying with community users.

metaci/cumulusci/views.py
```diff
@@ -110,10 +114,16 @@ def get_org_config(org):
...
 def org_instance_delete(request, org_id, instance_id):
     instance = get_object_or_404(ScratchOrgInstance, org_id=org_id, id=instance_id)
 
+    # Verify access
+    try:
+        org = Org.objects.for_user(request.user).get(id=org_id)
+    except Org.DoesNotExist:
+        return HttpResponseForbidden('You are not authorized to view this org')
```
cdcarter: PermissionDenied...but maybe just 404?

metaci/cumulusci/views.py
```diff
@@ -123,10 +133,17 @@ def org_instance_delete(request, org_id, instance_id):
...
-@staff_member_required
+@login_required
 def org_instance_detail(request, org_id, instance_id):
     instance = get_object_or_404(ScratchOrgInstance, org_id=org_id, id=instance_id)
+
+    # Verify access
+    try:
+        org = Org.objects.for_user(request.user).get(id=org_id)
+    except Org.DoesNotExist:
+        return HttpResponseForbidden('You are not authorized to view this org')
```
cdcarter: permissiondenied

metaci/cumulusci/views.py
```diff
@@ -123,10 +133,17 @@ def org_instance_delete(request, org_id, instance_id):
...
 
-
-@staff_member_required
+@login_required
 def org_instance_detail(request, org_id, instance_id):
     instance = get_object_or_404(ScratchOrgInstance, org_id=org_id, id=instance_id)
+
+    # Verify access
+    try:
+        org = Org.objects.for_user(request.user).get(id=org_id)
```
cdcarter: is there a reason we aren't just doing `request.user.has_perms('org_login', instance.org)` ?

metaci/plan/migrations/0022_migrate_to_groups.py
```diff
@@ -0,0 +1,98 @@
...
+    for user in users.iterator():
+        user.groups.add(
+            run_group,
+            view_group,
+            rebuild_group,
+            qa_group,
+            org_group,
+        )
+        # Since access is now granted through guardian, remove is_staff
+        user.is_staff = False
```
cdcarter: should we leave `is_staff` for folks with superuser?

metaci/plan/views.py
```diff
@@ -35,33 +31,46 @@ def plan_detail(request, plan_id):
...
+        'repo': planrepo.repo,
     }
     return render(request, 'plan/plan_repo_detail.html', context=context)
 
-@staff_member_required
 def plan_run(request, plan_id):
     plan = get_object_or_404(Plan, id=plan_id)
-    context = {'plan': plan, 'planrepos': plan.planrepository_set.should_run().all()}
+    if not Plan.objects.for_user(request.user, 'plan.run_plan').filter(id=plan_id).count():
+        return HttpResponseForbidden('You are not authorized to run this plan')
```
cdcarter: permissiondenied

metaci/plan/views.py
```diff
@@ -35,33 +31,46 @@ def plan_detail(request, plan_id):
...
-    planrepo = get_object_or_404(PlanRepository, plan_id=plan.id, repo_id=repo.id, active=True, plan__active=True)
+    planrepo = get_object_or_404(
+        PlanRepository,
+        plan_id=plan.id,
+        repo_id=repo.id,
+        active=True,
+        plan__active=True
+    )
+    if not request.user.has_perm('plan.run_plan', planrepo):
+        return HttpResponseForbidden('You are not authorized to run this plan')
```
cdcarter: permissiondenied

metaci/build/views.py
```diff
@@ -29,21 +29,13 @@ def build_list(request):
...
-            return HttpResponseForbidden(
-                'You are not authorized to view this build')
-        if tab == 'org':
-            return HttpResponseForbidden(
-                "You are not authorized to view this build's org info")
-        if tab == 'qa':
-            return HttpResponseForbidden(
-                "You are not authorized to view this build's QA info")
+    if not request.user.has_perm('plan.view_builds', build.planrepo):
+        return HttpResponseForbidden(
```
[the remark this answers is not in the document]
jlantz: I honestly didn't know that's how it worked, will change :)

metaci/cumulusci/views.py
jlantz: Because the user doesn't have org_login on the org itself.  They inherit that perm from the PlanRepository that's using the org.

metaci/plan/migrations/0022_migrate_to_groups.py
jlantz: is_superuser implies is_staff inherently

metaci/cumulusci/views.py
cdcarter: aha, I think that might come to be confusing later on (which is probably what @davisagli was getting at in his other comment). this is good for now but we probably want to move to permissions on org also, since PlanRepository and Org have a ...strange relationship, to say the least.

metaci/hirefire/views.py
```diff
@@ -29,7 +29,7 @@ def info(request, token):
     if not settings.HIREFIRE_TOKEN:
         return HttpResponseBadRequest('Hirefire not configured.  Set the HIREFIRE_TOKEN environment variable on the app to use Hirefire for dyno scaling')
     if token != settings.HIREFIRE_TOKEN:
-        return HttpResponseForbidden('Invalid token')
```
cdcarter: This one I think is actually fine, since it's going to end up in a log somewhere on hirefire, we don't need the full template

metaci/build/views.py
```diff
@@ -77,51 +69,88 @@ def build_detail(request, build_id, rebuild_id=None, tab=None):
...
                 outcome__in=['Fail', 'CompileFail'])))
 
-    if rebuild and rebuild.org_instance:
-        org_instance = rebuild.org_instance
-    else:
-        org_instance = build.org_instance
-    context = {
+    obj_perms = {
+        'rebuild_builds': request.user.has_perm('plan.rebuild_builds', build.planrepo),
+        'org_login': request.user.has_perm('plan.rebuild_builds', build.planrepo),
```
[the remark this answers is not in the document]
jlantz: Fixed in latest commit

PR #307, opened 2018-10-20
Fix migration issues and use Public group for shared anonymous and logged in perms

jlantz:
- Make Build.planrepo null=True to handle migrations where no PlanRepository exists for build.  Builds without a planrepo are not accessible in the front end
- Change from granting AnonymousUser permissions directly and instead use the group Public to represent permissions for all users
- Add a signal on user creation to add the user to the Public group
- Add migration logic to add all existing users to the Public group


metaci/build/migrations/0019_build_planrepo.py
```diff
@@ -8,7 +8,11 @@
 def set_planrepository(apps, schema_editor):
     Build = apps.get_model('build', 'Build')
     for build in Build.objects.all().iterator():
-        build.planrepo = build.plan.planrepository_set.get(repo = build.repo)
+        try:
+            build.planrepo = build.plan.planrepository_set.get(repo = build.repo)
+        except Exception as e:
```
cdcarter: you should be able to catch ObjectDoesNotExist instead, which is the base of all Model.DoesNotExist exceptions.

metaci/build/models.py
```diff
@@ -117,7 +117,7 @@ class Build(models.Model):
     tag = models.CharField(max_length=255, null=True, blank=True)
     pr = models.IntegerField(null=True, blank=True)
     plan = models.ForeignKey('plan.Plan', related_name='builds', on_delete=models.CASCADE)
-    planrepo = models.ForeignKey('plan.PlanRepository', related_name='builds', on_delete=models.CASCADE)
+    planrepo = models.ForeignKey('plan.PlanRepository', related_name='builds', on_delete=models.CASCADE, null=True)
```
cdcarter: :tada:

metaci/plan/migrations/0022_migrate_to_groups.py
```diff
@@ -64,7 +62,15 @@ def create_groups_and_migrate_users(apps, schema_editor):
...
+    Permission = apps.get_model('auth.Permission')
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    GroupObjectPermission = apps.get_model('guardian.GroupObjectPermission')
+
+    public_group = Group.objects.create(name='Public')
+    view_builds = Permission.objects.get(codename='view_builds')
+
     try:
         anon = get_anonymous_user()
     except Exception as e:
```
cdcarter: can just catch User.DoesNotExist ?

metaci/plan/migrations/0022_migrate_to_groups.py
```diff
@@ -75,12 +81,37 @@ def grant_anonymous_perms(apps, schema_editor):
             password=make_password(None),
         )
         anon.save()
+
+    try:
+        anon.groups.get(name='Public')
+    except Exception as e:
```
cdcarter: should be able to just catch Group.DoesNotExist...

but maybe consider:
```
if not anon.groups.filter(name='Public').exists():
    anon.groups.add(public_group)
```

metaci/plan/migrations/0022_migrate_to_groups.py
```diff
@@ -75,12 +81,37 @@ def grant_anonymous_perms(apps, schema_editor):
...
+            content_type = content_type,
+            object_pk = pr.id,
+        )
+        obj_perm.save()
+
+    # Add all users to the Public group
+    for user in User.objects.all().iterator():
+        try:
+            user.groups.get(name='Public')
+        except Exception as e:
```
cdcarter: same change as on anon user.

metaci/users/handlers.py
```diff
@@ -0,0 +1,22 @@
...
+    user = kwargs['instance']
+    created = kwargs['created']
+    if not created:
+        return
+
+    if user.username == guardian_settings.ANONYMOUS_USER_NAME:
+        return
+
+    try:
+        user.groups.get(name='Public')
```
cdcarter: is there a scenario where in the initial post_save, the user already has the group assigned? this seems extraneous, since we've already short circuited if this wasn't the save where the user was created.

metaci/plan/migrations/0022_migrate_to_groups.py
jlantz: Yeah, migrations are kinda the twilight zone when it comes to models.  The error that's thrown is actually metaci.users.models.DoesNotExist but if you try to build an except using that path you get a ModuleError since that doesn't exist.  There's some real "magic" in how migrations work which was a PITA when figuring this stuff out, but I've tested this against a clone of the mrbelvedereci-staging database and against a clean database.

cdcarter: I *think* `ObjectDoesNotExist` should catch it (And any DoesNotExist-ish errors), but yea, if this is what it takes.

jlantz: It does, pushed new commit using ObjectDoesNotExist

metaci/plan/migrations/0022_migrate_to_groups.py
```diff
@@ -64,23 +63,46 @@ def create_groups_and_migrate_users(apps, schema_editor):
...
         if pr.plan.public is False:
             continue
         if pr.repo.public is False:
             continue
-        assign_perm('plan.view_builds', anon, pr)
+
+        obj_perm = GroupObjectPermission(
+            permission=view_builds,
+            group = public_group,
+            content_type = content_type,
```
cdcarter: If this is a GenericForeignKey (I think it is?) this should be the ContentType for the object (in this case PlanRepository)

cdcarter: ...and it looks like its the ContentType for GroupObjectPermission itself, meaning the GFK will have confusion (at best)

metaci/users/handlers.py
```diff
@@ -0,0 +1,19 @@
...
+    user = kwargs['instance']
+    created = kwargs['created']
+    if not created:
+        return
+
+    if user.username == guardian_settings.ANONYMOUS_USER_NAME:
+        return
+
+    user.groups.add(
+        Group.objects.get(name='Public')
```
cdcarter: Kinda wondering if we shouldn't put the Public group name in a setting (like the Anon user is...) Meh, nbd, just a thought for other installations

PR #324, opened 2018-11-14
Fix release creation migration script

jlantz:
- Added metaci.release.utils.update_release_from_github
- Added method Release.update_from_github
- Changed migration script to create release then call
  update_release_from_github to avoid Github API throttling caused by

### DIFF
--- a/metaci/release/migrations/0003_populate_releases.py
+++ b/metaci/release/migrations/0003_populate_releases.py
@@ -31,6 +31,13 @@ def populate_releases(apps, schema_editor):
         repo = github.repository(info['repo'].owner, info['repo'].name) 
         for release in repo.iter_releases():
             if release.tag_name in info['tags']:
+                existing = Release.objects.filter(
+                    repo = info['repo'],
+                    git_tag = release.tag_name,
+                )
+                if existing.exists():
+                    continue
+                
                 ref = repo.ref('tags/{}'.format(release.tag_name))
                 rel = Release(
                     repo = info['repo'],
--- a/metaci/release/migrations/0003_populate_releases.py
+++ b/metaci/release/migrations/0003_populate_releases.py
@@ -11,6 +11,8 @@ from django.db import migrations
 from django.utils.dateparse import parse_date
 from github3 import login
 
+from metaci.release.utils import update_release_from_github
+
 logger = logging.getLogger(__name__)
 
 def populate_releases(apps, schema_editor):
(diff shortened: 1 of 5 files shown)


jlantz:
NOTE: I originally had all the logic for Github interaction in Release.update_from_github but had to move to a utility so it was accessible from the migration script.  Migration scripts can't access model methods.

metaci/release/models.py
```diff
@@ -1,13 +1,23 @@
...
 from django.utils.translation import ugettext_lazy as _
 
 from model_utils import Choices
 from model_utils.models import StatusModel
 from model_utils.fields import AutoCreatedField, AutoLastModifiedField
 
+from metaci.release.utils import update_release_from_github
+from metaci.repository.utils import get_github_api
+
+logger = logging.getLogger(__name__)
```
davisagli: Looks like you added imports and a logger which can be removed now that the code was moved to utils.

metaci/release/migrations/0003_populate_releases.py
```diff
@@ -29,49 +31,21 @@ def populate_releases(apps, schema_editor):
...
-                    rel.package_version_id = package_version_id[0]
-
-                trialforce_id = re.findall(
-                    r'^(0TT[\w]{15,18})', release.body
-                )
-                if trialforce_id:
-                    rel.trialforce_id = trialforce_id[0]
-
-                rel.save()
+        for tag in info['tags']:
```
davisagli: How many releases need to get created in production? Is it still enough to cause problems with the github API?

metaci/repository/utils.py
```diff
@@ -1,14 +1,18 @@
 from django.conf import settings
 from github3 import login
 
+def get_github_api(repo):
+    github = login(settings.GITHUB_USERNAME, settings.GITHUB_PASSWORD)
```
davisagli: If you use `cumulusci.core.github.get_github_api` it'll set up a `requests` session that retries on transient errors.

metaci/release/migrations/0003_populate_releases.py
jlantz: There's 88 releases to be created and the limit is 5000 per hour so we should be good.  There are a bunch of releases in our repositories so the previous implementation which iterated through all releases was the issue.

metaci/repository/utils.py
jlantz: I think that will work given that the service is in the db.  I'll test it out locally and push if it works.  It would mean the migration script would fail if you don't have the Github service configured in MetaCI vs using the environment variables which are required to be set per validation logic in settings.py

davisagli: I'm not following. You still pass the username and password in.

cdcarter: (`get_github_api` is a helper that just gets a github3 client that autoretries, it doesn't touch the keychain)

david - we should likely move it into cumulusci.utils since it doesnt at all rely on a config/runtime/context

cdcarter: the exact incantation would be 
```
github = get_github_api(settings.GITHUB_USERNAME, settings.GITHUB_PASSWORD)
return github.repository(repo.owner, repo.name)
```

it just provides some extra nicenes

metaci/release/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+    package_version_id = re.findall(r'(04t[\w]{15,18})', gh_release['body'])
+    if package_version_id:
+        release.package_version_id = package_version_id[0]
+
+    trialforce_id = re.findall(
+        r'^(0TT[\w]{15,18})', gh_release['body']
+    )
+    if trialforce_id:
+        release.trialforce_id = trialforce_id[0]
```
cdcarter: i'd `return release` here, so that we can fluently call `update_release_from_github(my_release).save()`

cdcarter: also backlog item (how do we call it? ;) ) but:
```
@rq.job(queue="short")
def update_release(rel_id: Int) -> bool:
   release = Release.objects.get(pk=rel_id)
   update_release_from_github(release)
   release.save()
```

metaci/release/models.py
```diff
@@ -1,13 +1,20 @@
...
+from django.utils.dateparse import parse_date
+from django.utils.dateparse import parse_datetime
 from django.utils.translation import ugettext_lazy as _
 
 from model_utils import Choices
 from model_utils.models import StatusModel
 from model_utils.fields import AutoCreatedField, AutoLastModifiedField
 
+from metaci.release.utils import update_release_from_github
+from metaci.repository.utils import get_github_api
```
davisagli: @jlantz There are still some unused imports. Install pylint in your virtualenv and VSCode should show errors about them.

metaci/repository/utils.py
```diff
@@ -1,14 +1,16 @@
 from django.conf import settings
-from github3 import login
+from cumulusci.core.github import get_github_api
 
+def get_github_api(repo):
+    github = get_github_api(settings.GITHUB_USERNAME, settings.GITHUB_PASSWORD)
```
davisagli: I think you'd better rename the local `get_github_api` function, because I'm not sure that this won't call itself.

jlantz: I opted to remove repository.utils.get_github_api in favor of the existing Repository.github_api method, but this did require manually constructing the repo api in migrations which can't access that method.  Tested running migration locally.

PR #325, opened 2018-11-15
Fix perm check for QA tab

jlantz:


davisagli:
@jlantz can you fix the repo plan view while you're at it? e.g. https://mrbelvedereci.herokuapp.com/plans/75/SalesforceFoundation/Cumulus

PR #333, opened 2018-11-28
(WIP) Port to Python 3

davisagli:


cdcarter:
TODO:

- [ ] django-defender / redis3 incompatibility
- [ ] jwt signer/verifier fns has been deprecated, likely need to update jwt lib
- [ ] test with ccilmatest

PR #335, opened 2018-12-05
Fix triggering builds from new tags

davisagli:
(Github appears to now be only sending a single webhook for new tags,
probably due to changes in how tags are created related to the github3.py upgrade)


davisagli:
I'm going to go ahead and merge this so I can deploy and test it while builds are still quiet.

PR #338, opened 2018-12-17
Store output.xml as BuildFlow asset

davisagli:
So far I still haven't been able to figure out what's happening with builds like these where Robot Framework reports a failure but all the individual tests passed. I'm hoping we can spot what's going on by looking at the raw output.xml


metaci/testresults/importer.py
```diff
@@ -135,9 +135,18 @@ def populate_limit_fields(testresult, code_unit):
 
 
 def import_robot_test_results(build_flow, path):
+    # import is here to avoid import cycle
+    from metaci.build.models import BuildFlowAsset
```
cdcarter: "next time" i'd use apps.get_model().

PR #340, opened 2018-12-18
Check for DevHub Capacity

cdcarter:
In check_queued_build, check the number of remaining scratch orgs before trying the build.

This can definitely be done cleaner... Just wanted to get an idea out there for comment...

should we allow MetaCI to use all orgs, or is reserving a capacity (i did ten in this version) useful? Obviously this isn't a strict guarantee and there are race conditions to consider...

Should we cache the number available in some way, so that we don't hit the devhub with another call on every queued build? Probably not that big of a deal and would invalidate quickly.

This requires functional testing on staging, to test how it behaves with the full check_waiting_builds lifecycle. I should probably seperate the gate from the requeueing logic, so that on staging we can fake low capacity.

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -290,6 +290,8 @@ SFDX_HUB_USERNAME = None
 # Application Behaviors
 GITHUB_STATUS_UPDATES_ENABLED = env.bool("GITHUB_STATUS_UPDATES_ENABLED", True)
 METACI_FLOW_SUBCLASS_ENABLED = env.bool("METACI_FLOW_SUBCLASS_ENABLED", True)
+SCRATCH_ORG_RESERVE = env.int("METACI_SCRATCH_ORG_RESERVE", 10)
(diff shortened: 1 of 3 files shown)


config/settings/common.py
```diff
@@ -290,6 +290,8 @@
 # Application Behaviors
 GITHUB_STATUS_UPDATES_ENABLED = env.bool("GITHUB_STATUS_UPDATES_ENABLED", True)
 METACI_FLOW_SUBCLASS_ENABLED = env.bool("METACI_FLOW_SUBCLASS_ENABLED", True)
+SCRATCH_ORG_RESERVE = env.int("METACI_SCRATCH_ORG_RESERVE", 10)
+# number of scratch orgs to leave available in the org.
```
davisagli: By default, this won't allow builds to run with a DE dev hub.

cdcarter: good catch, lemme set a very low default in the app.json for free setup.

PR #362, opened 2019-01-18
Update to use FlowCoordinator

davisagli:
I haven't tested this yet but this is more or less the shape of it.


config/settings/common.py
```diff
@@ -288,7 +288,7 @@
 
 # Application Behaviors
 GITHUB_STATUS_UPDATES_ENABLED = env.bool("GITHUB_STATUS_UPDATES_ENABLED", True)
-METACI_FLOW_SUBCLASS_ENABLED = env.bool("METACI_FLOW_SUBCLASS_ENABLED", True)
+METACI_FLOW_CALLBACK_ENABLED = env.bool("METACI_FLOW_SUBCLASS_ENABLED", True)
```
davisagli: Intentionally left the env var unchanged for backwards compatibility, not sure if we're using it.

metaci/build/models.py
```diff
@@ -534,26 +536,16 @@ def run_flow(self, project_config, org_config):
             raise FlowNotFoundError('Flow not found: {}'.format(self.flow))
         flow_config = FlowConfig(flow)
 
-        if settings.METACI_FLOW_SUBCLASS_ENABLED:
-            class_path = 'metaci.build.flows.MetaCIFlow'
-        else:
-            # Get the class to look up options
-            class_path = flow_config.config.get('class_path',
-                                                'cumulusci.core.flows.BaseFlow')
```
davisagli: Is it okay to drop this?

cdcarter: yep, we don't support flow_config.class_path anymore.

config/settings/common.py
cdcarter: go ahead and change the var too. the current impl is broken and off in all environments.

metaci/build/flows.py
```diff
@@ -1,17 +1,16 @@
...
 
-class MetaCIFlow(BaseFlow):
-    """ An implementation of BaseFlow that logs task execution to the database. """
+class MetaCIFlowCallback(FlowCallback):
+    """ An implementation of FlowCallback that logs task execution to the database. """
 
-    def __init__(self, *args, **kwargs):
-        self.buildflow_id = None
-        super(MetaCIFlow, self).__init__(*args, **kwargs)
+    def __init__(self, buildflow_id):
```
cdcarter: can just use the default constructor and the ctx variable?

config/settings/common.py
davisagli: Are you saying change the var here and in all the environments that turn it off, or change it here only with a new default of False?

metaci/build/flows.py
```diff
@@ -1,17 +1,16 @@
...
+        self.buildflow_id = buildflow_id
     
-    def _pre_task(self, task):
+    def pre_task(self, task):
```
cdcarter: the first arg to these is now the StepSpec, not the BaseTask, so, step.path, step.step_num etc...

metaci/build/models.py
```diff
@@ -534,26 +536,16 @@ def run_flow(self, project_config, org_config):
...
-
-        flow_class = import_class(class_path)
+        callbacks = None
```
cdcarter: there aren't checks for if callbacks is none in teh flowrunner (though maybe there should be) so this needs to be a FlowCallback if its not a MetaCIFlowCallback

metaci/build/models.py
```diff
@@ -534,26 +536,16 @@ def run_flow(self, project_config, org_config):
...
+        if settings.METACI_FLOW_CALLBACK_ENABLED:
+            callbacks = MetaCIFlowCallback(buildflow_id=self.pk)
```
cdcarter: again switch to using default context constructor

metaci/build/flows.py
davisagli: I think this makes it more readable by making it obvious what the context _is_.

metaci/build/flows.py
davisagli: Hmm. I can mostly work with that but it means we don't have the final initialized options.

metaci/build/models.py
davisagli: There is:

        if not callbacks:
            callbacks = FlowCallback()

config/settings/common.py
cdcarter: i'm saying 
```
METACI_FLOW_CALLBACK_ENABLED = env.bool("METACI_FLOW_CALLBACK_ENABLED", True)
```
and we can set METACI_FLOW_CALLBACK_ENABLED to False after deploying if we run into an issue.

This is meant to be a killswitch. the old impl was bad because of how it handled subflows. the flowcoordinator gets rid of that so lets plan on it working.

metaci/build/flows.py
cdcarter: i mean, we can change it, but thats what the code does now so this will fail wth attributeerrors.

hmm.

metaci/build/models.py
cdcarter: oh. heh.

metaci/build/flows.py
cdcarter: okay sure. can you change the cci branch then to not have the default constructor/ctx member?

davisagli: Done

PR #367, opened 2019-02-04
Drop indexes for most columns on testresults table

jlantz:


davisagli:
I'll plan to merge and deploy this this evening if there's a gap in builds.

PR #370, opened 2019-02-06
PlanRepositoryTrigger

jstvz:
This PR adds the `PlanRepositoryTrigger` (PRT) model and associated handler for the `build_complete` signal. The handler looks up any PRTs for the completed build's `planrepo` and calls each instance's `fire()` method to enqueue the target build.


metaci/plan/models.py
```diff
@@ -205,3 +206,49 @@ def get_absolute_url(self):
...
+    class Meta:
+        ordering = ["target_plan_repo", "source_plan_repo"]
+        unique_together = ("target_plan_repo", "source_plan_repo")
+        verbose_name_plural = "Plan Repository Triggers"
+        permissions = (
+            ("run_plan", "Run Plan"),
+            ("view_builds", "View Builds"),
+            ("rebuild_builds", "Rebuild Builds"),
+            ("org_login", "Login to Org"),
+        )
```
davisagli: Are these permissions actually relevant here?

metaci/plan/models.py
```diff
@@ -205,3 +206,49 @@ def get_absolute_url(self):
...
+        PlanRepository, on_delete=models.CASCADE, related_name="Triggers"
+    )
+    branch = models.CharField(max_length=255)
+    active = models.BooleanField(default=True)
+
+    objects = PlanRepositoryTriggerQuerySet.as_manager()
+
+    class Meta:
+        ordering = ["target_plan_repo", "source_plan_repo"]
+        unique_together = ("target_plan_repo", "source_plan_repo")
```
davisagli: Hmm. What if we want to trigger builds for multiple branches in the target repository?

metaci/plan/handlers.py
```diff
@@ -0,0 +1,16 @@
+from django.dispatch import receiver
+from metaci.build.signals import build_complete
+from metaci.plan.models import PlanRepositoryTrigger
+
+
+@receiver(build_complete)
+def trigger_dependent_builds(sender, **kwargs):
+    build = kwargs.get("build")
+    status = kwargs.get("status")
```
davisagli: Any reason not to just make the function signature be `(sender, build, status)`?

metaci/plan/models.py
```diff
@@ -205,3 +206,49 @@ def get_absolute_url(self):
...
+            repo=self.target_plan_repo.repo, name=self.branch
+        )
+        return branch
+
+    def fire(self, finished_build):
+        build = Build(
+            repo=self.target_plan_repo.repo,
+            plan=self.target_plan_repo.plan,
+            planrepo=self.target_plan_repo,
+            commit=self.branch,
```
davisagli: If the new build doesn't start running immediately, do we want it to use the state of this branch from the time the build was queued or the time it runs? If the former, we should resolve this to an actual commit sha like the plan form does in metaci/plan/forms.py

metaci/plan/models.py
jstvz: Cargo-culted a bit here. Will remove.

metaci/plan/models.py
jstvz: Good question. I hadn't considered the use case. I can add `branch` to `unique_together`.

metaci/plan/handlers.py
jstvz: My understanding is that it's less tightly coupled to the `build_complete` signal's arguments. ([docs](https://docs.djangoproject.com/en/2.1/topics/signals/#receiver-functions))

metaci/plan/models.py
jstvz: I was operating under the assumption that it was the latter, but resolving to an actual hash would be consistent with our behavior elsewhere, so I'll update.

PR #378, opened 2019-03-02
Django filter renamed the name= parameter to field_name= .

prescod:
https://django-filter.readthedocs.io/en/master/guide/migration.html#filter-name-renamed-to-filter-field-name-792

Fixes crashes on pages like this:

https://mrbelvedereci.herokuapp.com/tests/60297/ci_feature/compare-to


davisagli:
@prescod there are some changes in metaci/testresults/models.py that go beyond renaming name to field_name -- are those intended to be here too?

PR #380, opened 2019-03-06
Feature/perf page

prescod:
Work in progress of a performance review page.

REST API draft is done.


metaci/api/views/testmethod_perf.py
```diff
@@ -0,0 +1,284 @@
...
+        widget=DateRangeWidget(attrs={"type": "date"}),
+    )
+
+    # This is not really a filter. It's actually just a query input but putting it
+    # here lets me get it in the form.
+    build_flows_limit = django_filters.rest_framework.NumberFilter(
+        method="dummy_filter", label="Build Flows Limit (default: 100)"
+    )
+
+    disable_by_default = dir()  # disable all of these fields by default.
```
davisagli: I didn't realize that `dir()` could be used without an argument to list the names of locals, so thanks for the comment. I'm not entirely convinced it's better than being explicit.

metaci/api/views/testmethod_perf.py
```diff
@@ -0,0 +1,284 @@
...
+    filter_class = TestMethodPerfFilter
+
+    # example URLs:
+    # http://localhost:8000/api/testmethod_perf/?repo=gem&plan=Release%20Test&method_name=testCreateNegative
+    # http://localhost:8000/api/testmethod_perf/?repo=Cumulus&plan=Feature%20Test&o=avg
+    # http://localhost:8000/api/testmethod_perf/?method_name=&repo=&plan=&flow=&recentdate=&daterange_after=&daterange_before=&o=-repo
+
+    def _get_buildflows(self):
+        """Which buildflows do we need to look at? Limit by time, repo, etc.
+           Also limit # for performance reasons"""
```
davisagli: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings recommends indenting subsequent docstring lines to the same level as the quotes on the first line, and putting the closing quotation marks on their own line.

metaci/api/views/testmethod_perf.py
```diff
@@ -0,0 +1,284 @@
...
+        for fieldname in fields_to_include:
+            fieldname = fieldname.strip("-")
+            aggregations[fieldname] = fields[fieldname]
+
+        print("Aggregating", aggregations)
+
+        return aggregations
+
+    def _get_splitter_fields(self):
+        "Which fields to split on (or group by, depending on how you think about it"
```
davisagli: Please use triple-quotes for consistency. Makes it more obvious at a glance that it's the docstring.

metaci/api/views/testmethod_perf.py
```diff
@@ -0,0 +1,284 @@
...
+        params = self.request.query_params
+        output_fields = {}
+
+        build_fields = BuildFlowFilterSet.build_fields
+        if params.get("group_by"):
+            for param in params.getlist("group_by"):
+                output_fields[param] = F("build_flow__" + build_fields[param])
+
+        print("Splitting", output_fields)
```
davisagli: Please use a logger

metaci/api/views/testmethod_perf.py
```diff
@@ -0,0 +1,284 @@
...
+
+class TurnFilterSetOffByDefaultBase(django_filters.rest_framework.FilterSet):
+    """A bit of a hack to allow two filtersets to work together in generating the
+       djago-filter config form but to have only one of them actually filter the
+       output queryset. The other filters a queryset of a sub-select."""
+
+    really_filter = False
+
+    def __init__(self, *args, really_filter=really_filter, **kwargs):
+        self.really_filter = really_filter
```
davisagli: `really_filter` is only used in this method; is there a reason to store it as an instance variable?

metaci/api/serializers/testmethod_perf.py
```diff
@@ -0,0 +1,15 @@
...
+    def to_representation(self, instance):
+        result = super(NonNullModelSerializer, self).to_representation(instance)
+        return OrderedDict(
+            [(key, result[key]) for key in result if result[key] is not None]
+        )
+
+
+class TestMethodPerfSerializer(serializers.Serializer):
+    def to_representation(self, instance):
+        return instance
```
davisagli: Does this actually work, or is it just a placeholder?

metaci/api/serializers/testmethod_perf.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework import serializers
+from collections import OrderedDict
+
+
+class NonNullModelSerializer(serializers.Serializer):
```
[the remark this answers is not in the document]
prescod: Was obsolete. Deleted now.

metaci/api/views/testmethod_perf.py
```diff
@@ -0,0 +1,284 @@
...
+        super().__init__(*args, **kwargs)
+
+        for filtername in self.filters.keys():
+            if not really_filter:
```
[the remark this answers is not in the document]
prescod: Yes, good catch. At one point it was supposed to have an else.

metaci/api/views/testmethod_perf.py
prescod: I'll try a decorator and see if that looks better. I do kind of want to preserve the property that adding a method to this class will "automatically" give it the conditional behaviour. But a decorator might be clearer about where and how the magic is happening.

metaci/api/views/testmethod_perf.py
prescod: Those are just transient while I hack at it. I want them to stick out like a sore thumb on purpose. But I'm done with them now (and can add them back in easily if the need arises again) so I'll delete them. I'll try to remember to label them "# TODO: Remove" for WIP code reviews in the future.

davisagli: I figured so, and probably should have just kept quiet since it's still a draft PR.

metaci/api/views/testmethod_perf.py
prescod: Another one of those things that made sense at one time but doesn't anymore.

metaci/api/serializers/testmethod_perf.py
prescod: Can you take a look at the file now and see if it is clearer please?

https://github.com/SFDO-Tooling/MetaCI/blob/06ee51b2b9625c4cbab2d86b343e7fc5aa6746b7/metaci/api/serializers/simple_dict_serializer.py

metaci/api/views/testmethod_perf.py
prescod: @davisagli What do you think about it as a decorator?

https://github.com/SFDO-Tooling/MetaCI/blob/06ee51b2b9625c4cbab2d86b343e7fc5aa6746b7/metaci/api/views/testmethod_perf.py#L59

metaci/api/serializers/testmethod_perf.py
davisagli: Yeah, the part I was missing is that the values method of a queryset makes it return a dict rather than model isntances.

metaci/api/views/test_testmethod_perf.py
```diff
@@ -0,0 +1,335 @@
+from metaci.conftest import StaffSuperuserFactory, TestResultFactory
```
davisagli: Can you turn metaci/api/tests.py into a package (its existing tests can go in metaci/api/tests/test_urls.py) and move this file into that package? That will better match the structure of tests elsewhere in this project and will make it easier to configure pytest to only pay attention to modules inside folders named `tests`.

metaci/api/views/test_testmethod_perf.py
```diff
@@ -0,0 +1,335 @@
...
+        print(*args)  # Pytest does useful stuff with stdout, better than logger data
+
+    @classmethod
+    def make_user_and_client(cls):
+        user = StaffSuperuserFactory()
+        client = APIClient()
+        client.force_authenticate(user)
+        response = client.get("/api/")
+        if response.status_code == 400 and "DisallowedHost" in str(response.content):
+            cls.debugmsg("**** YOU MAY NEED TO ADD AN ALLOWED_HOSTS TO YOUR TEST.PY")
```
davisagli: I did add ALLOWED_HOSTS to the settings for tests, so I'm not sure we need this check now.

prescod: Fixed in: 

https://github.com/SFDO-Tooling/MetaCI/pull/380/commits/beaf7c37b45bc92fadabf61b899dfb0faaedd09b

metaci/api/views/test_testmethod_perf.py
prescod: Done in:

https://github.com/SFDO-Tooling/MetaCI/pull/380/commits/beaf7c37b45bc92fadabf61b899dfb0faaedd09b

PR #383, opened 2019-03-12
IP-based security for MetaCI APIs

prescod:
Should not be merged until the corresponding PR in sfdo-template-helpers has been merged and released.

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -11,6 +11,9 @@ https://docs.djangoproject.com/en/dev/ref/settings/
 from __future__ import absolute_import, unicode_literals
 
 import environ
+from ipaddress import IPv4Network
+from typing import List
+
 
 ROOT_DIR = (
     environ.Path(__file__) - 3
@@ -19,6 +22,19 @@ APPS_DIR = ROOT_DIR.path("metaci")
 
 env = environ.Env()
 env.read_env()
+
+
+def ipv4_networks(val: str) -> List[IPv4Network]:
+    return [IPv4Network(s.strip()) for s in val.split(",")]
+
+
+def url_prefix(val: str) -> str:
+    return val.rstrip("/") + "/"
+
+
+def url_prefix_list(val: str) -> List[str]:
+    return val.split(",")
+
 
 # APP CONFIGURATION
 # ------------------------------------------------------------------------------
@@ -80,6 +96,7 @@ MIDDLEWARE = (
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "sfdo_template_helpers.admin.middleware.AdminRestrictMiddleware",
 )
 
 # MIGRATIONS CONFIGURATION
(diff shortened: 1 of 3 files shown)


requirements/base.txt
```diff
@@ -77,3 +77,6 @@ PyJWT==1.7.1
 
 # Django Defender
 django-defender==0.6.0
+
+# SFDO Template Helpers for authorization middleware and other helpers
+-e git+https://github.com/SFDO-Tooling/sfdo-template-helpers.git@v0.1.0#egg=sfdo_template_helpers
```
davisagli: We don't currently release sfdo-template-helpers to PyPI, but once https://github.com/SFDO-Tooling/sfdo-template-helpers/pull/21 is merged I will push a new tag that we can refer to here.

config/settings/common.py
```diff
@@ -205,6 +222,19 @@
...
 
+# ADMIN_AREA_PREFIX = env("DJANGO_ADMIN_URL", default="admin/", cast=url_prefix)
+
+RESTRICTED_PREFIXES = env(
+    "RESTRICTED_PREFIXES", default=["admin/", "api/"], cast=url_prefix_list
+)
+
+ADMIN_API_ALLOWED_SUBNETS = env(
+    "ADMIN_API_ALLOWED_SUBNETS",
+    default="127.0.0.1/32",
```
davisagli: For backwards compatibility this should probably default to 0.0.0.0/0

We want to be more restrictive for our own deployments, but we don't want to catch other installations of MetaCI off guard.

config/settings/common.py
```diff
@@ -20,6 +23,19 @@
...
+def ipv4_networks(val: str) -> List[IPv4Network]:
+    return [IPv4Network(s.strip()) for s in val.split(",")]
+
+
+def url_prefix(val: str) -> str:
+    return val.rstrip("/") + "/"
+
+
+def url_prefix_list(val: str) -> List[str]:
+    return val.split(",")
```
davisagli: Is this supposed to call `url_prefix`?

config/settings/common.py
```diff
@@ -205,6 +222,19 @@
 # ------------------------------------------------------------------------------
 ROOT_URLCONF = "config.urls"
 
+# ADMIN_AREA_PREFIX = env("DJANGO_ADMIN_URL", default="admin/", cast=url_prefix)
+
+RESTRICTED_PREFIXES = env(
+    "RESTRICTED_PREFIXES", default=["admin/", "api/"], cast=url_prefix_list
```
davisagli: On line 273 we have:
```
ADMIN_URL = env("DJANGO_ADMIN_URL", default="admin")
ADMIN_URL_ROUTE = r"^{}/".format(ADMIN_URL)
```
and we do set this environment variable in production. ADMIN_URL_ROUTE is then used in urls.py to configure the admin route.

metadeploy does the same thing but called the setting ADMIN_AREA_PREFIX instead of ADMIN_URL_ROUTE. Which is why the middleware checks for the ADMIN_AREA_PREFIX setting. I think we should rename ADMIN_URL_ROUTE to ADMIN_AREA_PREFIX in MetaCI too, and then we can remove `admin/` from the defaults here.

requirements/base.txt
davisagli: The v0.6.0 tag now exists.

config/settings/common.py
prescod: Can you clarify whether you want to rename only the SETTINGS variable or also the Environment variable? And if the latter, wouldn't that be a breaking change?

davisagli: Yeah, just the setting

config/settings/common.py
prescod: Okay.

config/settings/common.py
prescod: Maybe an alias is safer for backwards compatibility? 

It is a settings file, after all, someone could theoretically have their own clone of production.py.

Maybe its just a theoretical rather than a practical concern?

davisagli: That's a fair point. Adding an alias sounds reasonable.

davisagli:
Appears to work great on staging. I'll deploy to production once no builds are running.

PR #385, opened 2019-03-13
Run tests with pytest instead of nose

davisagli:
Switching to pytest because nose-tap errors out on Python 3 when there's a test failure, and nose is generally unmaintained.

`./manage.py test` is no longer the recommended way to run the tests. Instead just run `pytest`.

Also adding a default ALLOWED_HOSTS setting that allows local access.

### DIFF
--- a/metaci/api/tests.py
+++ b/metaci/api/tests.py
@@ -1,7 +1,6 @@
 from rest_framework.reverse import reverse
 from test_plus.test import TestCase
 
-
 class TestURLReversals(TestCase):
     """Test URL patterns for APIs app."""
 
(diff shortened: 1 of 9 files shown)


davisagli:
@prescod note this option for reusing the test db https://pytest-django.readthedocs.io/en/latest/database.html#reuse-db-reuse-the-testing-database-between-test-runs

metaci/api/tests.py
```diff
@@ -1,7 +1,6 @@
 from rest_framework.reverse import reverse
 from test_plus.test import TestCase
 
-
```
prescod: I think Black is going to want to have a word with you.

PR #386, opened 2019-03-13
Quiet pytest warnings relating to certain classes.

prescod:
Pytest infers that any class with a name starting with the word Test must be a test-case. This checkin quiets warnings relating to certain classes.

Another option would be to disable all classname-based discovery and rely on subclass discovery only.

The pros/cons of that are documented here:

https://stackoverflow.com/a/43349055/11151197

Black also did its thing on some lines.

### DIFF
--- a/metaci/api/tests/test_testmethod_perf.py
+++ b/metaci/api/tests/test_testmethod_perf.py
@@ -1,4 +1,5 @@
 from metaci.conftest import StaffSuperuserFactory, TestResultFactory
+import pytest
 
 from urllib.parse import urlencode
 import random
@@ -49,6 +50,7 @@ class _TestingHelpers:
             TestResultFactory(**fields, build_flow__tests_total=1, method=t1.method)
 
 
+@pytest.mark.filterwarnings("ignore: Using the add method")
 class TestTestMethodPerfRESTAPI(APITestCase, _TestingHelpers):
     """Test the testmethodperf REST API"""
 
@@ -309,6 +311,7 @@ class TestTestMethodPerfRESTAPI(APITestCase, _TestingHelpers):
             self.assertIn(row["method_name"], ["Bar2", "Bar3"])
             self.assertNotIn(row["method_name"], ["Bar1", "Bar4"])
 
+    @pytest.mark.filterwarnings("ignore:DateTimeField")
     def test_filter_by_recent_date(self):
         yesterday = timezone.make_aware(datetime.today() - timedelta(1))
         day_before = timezone.make_aware(datetime.today() - timedelta(2))
--- a/metaci/api/views/testmethod_perf.py
+++ b/metaci/api/views/testmethod_perf.py
@@ -145,6 +145,8 @@ class BuildFlowFilterSet(TurnFilterSetOffByDefaultBase):
 class TestMethodPerfFilter(BuildFlowFilterSet, django_filters.rest_framework.FilterSet):
     """This filterset works on the output queries"""
 
+    __test__ = False  # Tell pytest that this is not a test.
+
     method_name = django_filters.rest_framework.CharFilter(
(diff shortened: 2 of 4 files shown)


jstvz:
Would one of the other answers work? Specifically:

```ini
[pytest]
python_files=test_*.py
```

...did the trick for me.

prescod:
> python_files=test_*.py

Seems totally reasonable and I was hoping to replace my checkin with a one-liner.

But I think it loads files and then looks through the entire Python objectspace whether the classes come from the loaded files or not. It is still picking up TestClass and TestMethod classes. I'm loading them because I'm testing them.

davisagli:
What are the warnings? My output is clean without these changes:
```
(metaci-py3) (pytest|✔) % pytest                                                                                                                           
========================================================================== test session starts ==========================================================================
platform darwin -- Python 3.6.5, pytest-4.3.0, py-1.7.0, pluggy-0.8.1
Django settings: config.settings.test (from ini file)
rootdir: /Users/dglick/Work/MetaCI, inifile: pytest.ini
plugins: tap-2.3, django-3.4.5, django-test-plus-1.1.1
collected 25 items                                                                                                                                                      

metaci/api/tests.py ..                                                                                                                                            [  8%]
metaci/plan/tests/test_handlers.py ..                                                                                                                             [ 16%]
metaci/plan/tests/test_models.py ......                                                                                                                           [ 40%]
metaci/users/tests/test_admin.py ..                                                                                                                               [ 48%]
metaci/users/tests/test_models.py ..                                                                                                                              [ 56%]
metaci/users/tests/test_urls.py ........                                                                                                                          [ 88%]
metaci/users/tests/test_views.py ...                                                                                                                              [100%]

======================================================================= 25 passed in 6.15 seconds =======================================================================
```

jstvz:
I'm confused, sorry. Is the problem that these classes are being discovered? If so, adding the underscore to `pytest.ini` works for me:

```
 λ pytest
======================================== test session starts =========================================
platform darwin -- Python 3.6.8, pytest-4.3.0, py-1.8.0, pluggy-0.9.0
Django settings: config.settings.test (from ini file)
rootdir: /Users/jestevez/projects/rel/MetaCI, inifile: pytest.ini
plugins: django-3.4.8, django-test-plus-1.1.1
collected 23 items

metaci/plan/tests/test_handlers.py ..                                                          [  8%]
metaci/plan/tests/test_models.py ......                                                        [ 34%]
metaci/users/tests/test_admin.py ..                                                            [ 43%]
metaci/users/tests/test_models.py ..                                                           [ 52%]
metaci/users/tests/test_urls.py ........                                                       [ 86%]
metaci/users/tests/test_views.py ...                                                           [100%]

===================================== 23 passed in 6.71 seconds ======================================
```

prescod:
Everybody else's output is clean without the warnings because nobody else is running code that imports those classes. But I have a branch with code that generates the warnings.

Why don't we hold this pull request until the other one is merged and then everyone can see it and we can experiment with techniques to fix it.

I'll just put up with the warnings for now.

jstvz:
Ah, now I get it. Thanks for the clarification.

davisagli:
I think we'll probably want to figure out how to restrict pytest to looking in tests directories, but it'll be easiest to do once we can reproduce the warnings.

prescod:
Okay now everyone should see the warnings on master.

If you have a node_modules directory you might also get an error which is fixed by adding this line to pytest.ini:

norecursedirs = node_modules dist src

https://www.reddit.com/r/ProgrammerHumor/comments/6s0wov/heaviest_objects_in_the_universe/

davisagli:
I just tested on master. The combination of `python_files=test_*.py` and `norecursedirs = node_modules dist src` appears to avoid the warnings, so I'd like to go with that rather than these markers on specific tests.

PR #387, opened 2019-03-15
Tests for perf page REST API

prescod:


metaci/api/views/test_testmethod_perf.py
```diff
@@ -0,0 +1,435 @@
...
+
+    email = factory.Sequence("user_{}@example.com".format)
+    username = factory.Sequence("user_{}@example.com".format)
+    password = factory.PostGenerationMethodCall("set_password", "foobar")
+    # socialaccount_set = factory.RelatedFactory(SocialAccountFactory, "user")
+
+
+class StaffSuperuserFactory(UserFactory):
+    is_staff = True
+    is_superuser = True
```
davisagli: In the long run, I think we'll want factories for all the models in a central place, rather than duplicating them in each test module that uses them. But we can wait and do that refactoring when we want to use them in a 2nd test module.

metaci/api/views/test_testmethod_perf.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def stats_test_helper(self, stat):
+        return self.api_call_helper(include_fields=stat)
+
+    def identical_tests_helper(self, count, method_name="GenericMethod", **fields):
+        t1 = TestResultFactory(
+            build_flow__tests_total=1, method__name=method_name, **fields
+        )
+        for i in range(count - 1):
+            TestResultFactory(**fields, build_flow__tests_total=1, method=t1.method)
```
davisagli: This is definitely a matter of taste but I don't find `helper` as part of a name to be very, well, helpful for reading what the at a glance what the function does without needing to look at its code. I would probably call these `get_api_results` and `insert_identical_test_results`, and (even more subjective feedback) move the `stats_test_helper` one-liner inline.

metaci/api/views/test_testmethod_perf.py
```diff
@@ -0,0 +1,435 @@
...
+class TestTestMethodPerfRESTAPI(APITestCase, _TestingHelpers):
+    """Test the testmethodperf REST API"""
+
+    @classmethod
+    def setUpClass(cls):
+        cls.client, cls.user = cls.make_user_and_client()
+        super().setUpClass()
+
+    def setUp(self):
+        self.client.force_authenticate(self.user)
```
davisagli: This already happens in make_user_and_client, no?

metaci/api/views/test_testmethod_perf.py
```diff
@@ -0,0 +1,435 @@
...
+            "success_percentage",
+        ]
+
+        def _test_fields(fields):
+            response = self.client.get(self.api_url(include_fields=fields))
+            self.assertEqual(response.status_code, 200)
+            rows = response.json()["results"]
+            for field in fields:
+                for row in rows:
+                    self.assertIn(field, row.keys())
```
davisagli: How about `self.assertSetEqual(set(fields), set(row))`

metaci/api/views/test_testmethod_perf.py
```diff
@@ -0,0 +1,435 @@
...
+            field = random.sample(includable_fields, 1)
+            _test_fields(field)
+
+        for i in range(10):
+            field1, field2 = random.sample(includable_fields, 2)
+            _test_fields([field1, field2])
+
+        for i in range(10):
+            field1, field2, field3 = random.sample(includable_fields, 3)
+            _test_fields([field1, field2, field3])
```
davisagli: What's the rationale for doing this random sampling with a preset seed?

davisagli:
Looks like the test run on Heroku CI is using sqlite. We should look into switching it to postgres using https://devcenter.heroku.com/articles/heroku-ci-in-dyno-databases

metaci/api/views/test_testmethod_perf.py
prescod: I agree. I actually did intend to move them but forgot before I checked in. Part of the reason that I didn't get around to it was because it wasn't obvious where to put them. Not sure if there is a convention for file naming for these factories.

We don't have a metaci/tests folder, we don't have a metaci/models folder and I'm not enthusiastic about breaking it up into a bunch of 5-line files.

davisagli: I would create metaci/conftest.py and put them in there. That'll match where they are in MetaDeploy, and is the pytest convention for where to put shared fixtures.

metaci/api/views/test_testmethod_perf.py
prescod: The rationale for random sampling is that I don't want to make all of the combinations by hand. (roughly 30 lines of code)

The rational for presetting the seed is that if you don't, the tests might fail intermittently every 3 years for a mysterious reason and you won't be able to debug it.

I had a "commute-thought" today that another way to do it would be to have the seed be the current month (or year), so you would get "new data" every month but you would have a whole month to debug the issue before it would go away on its own.

metaci/api/views/test_testmethod_perf.py
prescod: https://github.com/SFDO-Tooling/MetaCI/commit/158a21b382ca98e4e5c9422063e026b5bb715902

metaci/api/views/test_testmethod_perf.py
prescod: https://github.com/SFDO-Tooling/MetaCI/commit/89bd396f42daf6734c1369b1137ebdfdaba677f8

metaci/api/views/test_testmethod_perf.py
prescod: https://github.com/SFDO-Tooling/MetaCI/commit/19986e2

metaci/api/views/test_testmethod_perf.py
davisagli: I suppose the APIClient probably resets headers between tests. This is fine.

PR #390, opened 2019-03-19
Initial package of React/Django infrastructure.

prescod:
Too much to summarize! Basically the integration of sfdo-template and MetaCI.

One thing to discuss is whether we want Husky to take over our pre-commit hooks.


src/js/components/header/customDomainModal.js
```diff
@@ -0,0 +1,97 @@
...
+              .my.salesforce.com
+            </div>
+          </Input>
+        </form>
+      </Modal>
+    );
+  }
+}
+
+export default CustomDomainModal;
```
davisagli: I don't think we need this; it's related to Salesforce oauth login.

src/js/components/header/login.js
```diff
@@ -0,0 +1,136 @@
...
+          options={menuOpts}
+          onSelect={this.handleSelect}
+        />
+        <CustomDomainModal isOpen={modalOpen} toggleModal={this.toggleModal} />
+      </>
+    );
+  }
+}
+
+export default Login;
```
davisagli: We can drop this; it's for Salesforce oauth login.

yarn.lock
```diff
@@ -0,0 +1,14883 @@
...
+
+after@0.8.2:
+  version "0.8.2"
+  resolved "https://registry.yarnpkg.com/after/-/after-0.8.2.tgz#fedb394f9f0e02aa9768e702bda23b505fae7e1f"
+  integrity sha1-/ts5T58OAqqXaOcCvaI7UF+ufh8=
+
+ajv-errors@^1.0.0:
+  version "1.0.1"
+  resolved "https://registry.yarnpkg.com/ajv-errors/-/ajv-errors-1.0.1.tgz#f35986aceb91afadec4102fbd85014950cefa64d"
+  integrity sha512-DCRfO/4nQ+89p/RK43i8Ezd41EqdGIU4ld7nGF8OQ14oc/we5
```
davisagli: @prescod did you copy yarn.lock from sfdo-template or is this the result of running yarn after modifying package.json here?

prescod: I ran yarn.lock myself.

prescod: Yarn updates both package.json and yarn.lock. I only edit the requirements part of package.json by hand if some specific problem arises. I guess I trust that the Javascript workflow works for the Javascript world and the PIP workflow works for the Python world and I go with the, uh, flow.

PR #391, opened 2019-03-21
Add contributor docs

davisagli:
This starts to add contributor docs based on CONTRIBUTOR.rst in sfdo-template, along with a few adjustments to @prescod's work on incorporating the frontend infrastructure.

I have some more work to do on incorporating some of the material from the quip doc on MetaCI setup and making sure that the worker processes run correctly, but figured I'd submit this as a start.


CONTRIBUTING.rst
```diff
@@ -0,0 +1,158 @@
...
+
+.. _Python.org: https://www.python.org/downloads/
+
+There are a variety of tools that let you set up environment variables
+temporarily for a particular "environment" or directory. We use
+`virtualenvwrapper`_. Assuming you're in the repo root, do the following to
+create a virtualenv (once you have `virtualenvwrapper`_ installed locally)::
+
+    mkvirtualenv metaci --python=$(which python3.7)
+    setvirtualenvproject
```
davisagli: virtualenvwrapper is what OddBird uses but I don't know if any of us are using it at SFDO, so I'm open to recommending a different approach here.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,158 @@
...
+``locales/<language>/`` directories, and the `user language is auto-detected at
+runtime`_.
+
+During development, strings are parsed automatically from the JS, and an English
+translation file is auto-generated to ``locales_dev/en/translation.json`` on
+every build (``yarn build`` or ``yarn serve``). When this file changes,
+translations must be copied over to the ``locales/en/translation.json`` file in
+order to have any effect.
+
+Strings with dynamic content (i.e. known only at runtime) cannot be
```
prescod: Something doesn't seem quite clear to me here. I thought that only strings tagged with t() will be parsed. Otherwise every database value you read from the wire would cause a translation error.

CONTRIBUTING.rst
prescod: No strong preference.

CONTRIBUTING.rst
davisagli: Yes, `t()` is the marker for which strings are translated. This is just making a point about parsing that marker, that `t("literal string")` can be detected and added to the translation file at compile time, but `t(some_variable)` can't.

(This section is just copied verbatim from sfdo-template, fwiw.)

davisagli:
@prescod I think this now includes everything that's still relevant from the "Setting up MetaCI for local development" quip doc.

CONTRIBUTING.rst
jstvz: We made it clear there's more than one way to do it, so I think this is fine.

PR #394, opened 2019-03-28
Add SLDS to production STATICFILES_DIRS

davisagli:


config/settings/production.py
```diff
@@ -12,10 +12,9 @@
 """
 from __future__ import absolute_import, unicode_literals
 
-from django.utils import six
-
 import logging
 
+from django.utils import six
```
davisagli: I guess this was moved automatically by the isort precommit hook.

prescod: Oops, I didn't notice that this bit of "common.py" is not actually common. Arguably it should not go in the common file if it isn't shared? (not saying we should fix it today, but just speaking philosophically)

davisagli: It is different between them (dist vs dist/prod, to get the uglified webpack bundle I'm guessing)

PR #400, opened 2019-03-30
Add error page for vpn

prescod:


src/js/components/perfTable.js
```diff
@@ -243,6 +246,15 @@ export const UnwrappedPerfTable = ({doPerfRESTFetch, doPerfREST_UI_Fetch,
     </div>
   };
 
+const AuthError = () => {
+  return <main id="app" style={{ textAlign: "center" }} className="slds-grid__vertical slds-page-header">
+    <p className="slds-text-heading--medium">Please ensure you are on the VPN and logged in to view this data.</p>
+    <div style={{ paddingTop: "20px" }} className="slds-brand-band slds-brand-band_large">
+      <p><img src="https://i.gifer.com/G36W.gif" /></p>
```
jstvz: Is this a URL we control? If not, should it be?

prescod: Yes, you are exactly right. When I first put it in I wasn't sure whether it would get a good or bad reaction and I didn't want to bloat our git repository with something short-lived. Jason says that Kevin liked the image so I will pull it into git, but not in this PR. I want this one to be as small as possible for safety. I have a big refactoring branch for my component and I'll add the file there. Good catch!

PR #406, opened 2019-04-10
Bootstrap test data: First pass at this command. Not much data yet.

prescod:


metaci/conftest.py
```diff
@@ -18,7 +18,13 @@ class RepositoryFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = Repository
 
+    name = factory.LazyAttribute(
+        lambda a: "Repo_"
+        + factory.Faker("paragraph").generate({}).replace(" ", "_")[0:20].lower()
```
davisagli: How about the `word` faker so we don't cut off mid-word?

metaci/db/management/commands/populate_db.py
```diff
@@ -22,19 +31,36 @@ def handle(self, *args, **options):
...
+                Group.objects.get(name="Public"),
+                PublicPlanRepository,
+            )
+
+            fact.TestResultFactory(build_flow__build__planrepo=PublicPlanRepository)
+            fact.TestResultFactory(build_flow__build__planrepo=PrivatePlanRepository)
+            fact.TestResultFactory(build_flow__build__planrepo=PublicPlanRepository)
+            fact.TestResultFactory(build_flow__build__planrepo=PrivatePlanRepository)
+
+            self.make_consistent()
```
davisagli: I think you could use a `post_generation`-decorated method on TestResultFactory to increment the build_flow's `tests_total`. But it would involve more writes. This is fine.

prescod: Makes sense. Honestly, the only use I know of this db column is in some of my code that makes sure that it is greater than 0 to avoid looking at irrelevant build_flows. So it is probably not worth slowing down the test suite for. I probably should have just left it hard-coded to "5" as I did for the last month. ;)

metaci/conftest.py
prescod: I'll take a look.

PR #407, opened 2019-04-11
Object-level Security 

prescod:


metaci/permissions.py
```diff
@@ -0,0 +1,14 @@
+from django.core.exceptions import PermissionDenied, SuspiciousOperation
```
davisagli: Can you put this file in the metaci/api directory since it's for use with Django Rest Framework?

metaci/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+from rest_framework.permissions import BasePermission
+from sfdo_template_helpers.admin.middleware import AdminRestrictMiddleware
+
+
+class IsOnSecureNetwork(BasePermission):
+    def has_permission(self, request, view):
+        try:
+            # TODO: Should really rename _validate_ip to remove underscore.
```
davisagli: I think my preference would be to turn it into a standalone function that can be used both here and by AdminRestrictMiddleware. But we don't need to hold this up waiting for that.

PR #408, opened 2019-04-12
More test data than you can shake a stick at

prescod:
We've got

1.  a bit of documentation about [how to use it.](https://github.com/SFDO-Tooling/MetaCI/compare/feature/test-db?expand=1#diff-ea1f65a8eaac0cec7ae151dfdbbbc587)

2. more sophisticated [factories](https://github.com/SFDO-Tooling/MetaCI/compare/feature/test-db?expand=1#diff-199f1c9e48aa5ae11340b7a35588d9f2)

3. explicitly created [objects](https://github.com/SFDO-Tooling/MetaCI/compare/feature/test-db?expand=1#diff-27edd9fe29f15d32b1be7c2b9573b269)

4. fixes to the test cases [to deal with the changes](https://github.com/SFDO-Tooling/MetaCI/compare/feature/test-db?expand=1#diff-3d4d1f1e9c82a97cbdb38ec5a5867008

5. some method renaming for a testcase helper function because April Paul is smarter and nicer than March Paul.

### DIFF
--- a/metaci/conftest.py
+++ b/metaci/conftest.py
@@ -1,37 +1,89 @@
 import factory
 import factory.fuzzy
 
+import numbers
+import random
+
 from metaci.plan.models import Plan, PlanRepository
 from metaci.testresults.models import TestResult, TestMethod, TestClass
-from metaci.build.models import BuildFlow, Build
+from metaci.build.models import BuildFlow, Build, BUILD_STATUSES, BUILD_FLOW_STATUSES
 from metaci.repository.models import Branch, Repository
 
 from metaci.users.models import User
+
+BUILD_STATUS_NAMES = (name for (name, label) in BUILD_STATUSES)
+BUILD_FLOW_STATUS_NAMES = (name for (name, label) in BUILD_FLOW_STATUSES)
+
+
+def do_logs():
+    import logging
+
+    logger = logging.getLogger("factory")
+    logger.addHandler(logging.StreamHandler())
+    logger.setLevel(logging.DEBUG)
+
+
+def fake_name(prefix=None):
+    return factory.LazyAttribute(
+        lambda a: (getattr(a, "_name_prefix", None) or prefix or "")
+        + factory.Faker("word").generate({})
+    )
+
+
+class RelatedFactoryList(factory.RelatedFactory):
+    """Calls a factory 'size' times once the object has been generated.
+        Simplified from
+        https://github.com/FactoryBoy/factory_boy/blob/2481d411cde311bb7edd51b6dff3b345c2c14bc6/factory/declarations.py#L675
+        While awaiting release
+    """
+
+    def __init__(self, factory, factory_related_name="", size=2, **defaults):
+        self.size = size
+        super(RelatedFactoryList, self).__init__(
+            factory, factory_related_name, **defaults
+        )
+
+    def call(self, instance, step, context):
+        return [
+            super(RelatedFactoryList, self).call(instance, step, context)
+            for i in range(self.size if isinstance(self.size, int) else self.size())
+        ]
 
 
 class PlanFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = Plan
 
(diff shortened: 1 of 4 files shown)


metaci/conftest.py
```diff
@@ -1,51 +1,107 @@
...
+from metaci.build.models import BuildFlow, Build, BUILD_STATUSES, BUILD_FLOW_STATUSES
 from metaci.repository.models import Branch, Repository
 
 from metaci.users.models import User
 
+BUILD_STATUS_NAMES = (name for (name, label) in BUILD_STATUSES)
+BUILD_FLOW_STATUS_NAMES = (name for (name, label) in BUILD_FLOW_STATUSES)
+
+
+def do_logs():
```
davisagli: This looks unused?

metaci/conftest.py
```diff
@@ -1,51 +1,107 @@
...
+        lambda a: (getattr(a, "_name_prefix", None) or prefix or "")
+        + factory.Faker("word").generate({})
+    )
+
+
+class RelatedFactoryList(factory.RelatedFactory):
+    """Calls a factory 'size' times once the object has been generated.
+        Simplified from
+        https://github.com/FactoryBoy/factory_boy/blob/2481d411cde311bb7edd51b6dff3b345c2c14bc6/factory/declarations.py#L675
+        While awaiting release
```
davisagli: ?

metaci/conftest.py
```diff
@@ -1,51 +1,107 @@
...
         model = Plan
 
+    _name_prefix = "Plan"
+    name = fake_name()
+
 
 class RepositoryFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = Repository
+        exclude = ("_name_prefix",)
```
davisagli: Is there a reason not to just pass the prefix in as an argument to `fake_name`? That would avoid setting the `_name_prefix` attribute, so would avoid the need to exclude it here. But I suppose I'm missing something or else you would have done that...

metaci/conftest.py
prescod: I should add a comment to this effect:

"""Call this function at the module level to get more insight into what FactoryBoy is dong under the covers.""" 

Some future maintainer might need it and not know where to find it in the FactoryBoy docs.

metaci/conftest.py
prescod: Sorry to be unclear. I'll fix that comment too. The method is described in the docs for FactoryBoy but when I tried to use it, it wasn't there. It turns out that FactoryBoy has been awaiting release for about 10 months and the documentation is more up-to-date than the releases.

But that said: I ended up solving my problem another way, so I'll delete this class.

metaci/conftest.py
prescod: Maybe its overly theoretical, but it allows the constructor of the RepositoryFactory to override the name, which is in-line with the general style of FactoryBoy where you can push in stuff from outside. I'll experiment with it to ensure that it really works and really adds value in the data generated (e.g. labelling Public repos and Private repos to ensure that the private ones are hidden in the public view). If it doesn't add value or doesn't work right then I'll take it out.

davisagli: Ah okay, that makes sense. Though if it's for external use it should probably not start with an underscore.

prescod: https://github.com/SFDO-Tooling/MetaCI/pull/408/commits/1d3326acfe3421222ca126d5aa753580690dad6e

metaci/conftest.py
prescod: https://github.com/SFDO-Tooling/MetaCI/pull/408/commits/cb9ac02e086d411c208d84d35e2119e5e8efb7bd

metaci/conftest.py
prescod: https://github.com/SFDO-Tooling/MetaCI/pull/408/commits/cb9ac02e086d411c208d84d35e2119e5e8efb7bd#r275034674

PR #411, opened 2019-04-16
Clean error reporting

prescod:
The error message presumes that Raven will be working soon.

It looks like this:

image


static/images/broken.svg
```diff
@@ -0,0 +1,104 @@
+<svg id="broken" class="slds-illustration" viewBox="0 0 272 146" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
davisagli: According to https://lightningdesignsystem.com/downloads/ the SLDS images are licensed under https://creativecommons.org/licenses/by-nd/4.0/ so we probably need to link to that.

src/js/components/error.js
```diff
@@ -4,21 +4,23 @@ import * as React from 'react';
...
 
-class ErrorBoundary extends React.Component<Props, { hasError: boolean }> {
+class ErrorBoundary extends React.Component<
+  Props,
+  { hasError: boolean, info: {} },
+> {
   constructor(props: Props) {
     super(props);
-    this.state = { hasError: false };
+    this.state = { hasError: false, info: {} };
```
davisagli: Is the `info` state actually used somewhere?

src/js/components/error.js
```diff
@@ -27,9 +29,18 @@ class ErrorBoundary extends React.Component<Props, { hasError: boolean }> {
...
-            </Trans>
+            <>
+              <Trans i18nKey="anErrorOccurred">
+                Unless you pasted a URL incorrectly, it is probably a bug in the
+                code or a networking problem.
+                <br />
+                Our top minds have been alerted.
+                <br />
+                You may need to <a href="./perf">rebuild your query</a> from
+                scratch. Sorry about that.
```
davisagli: This is fine when we only have one React-based page, but I think we'll need a way to provide contextual instructions here eventually.

davisagli:
Also note the failing test.

static/images/broken.svg
prescod: Copyright added:

https://github.com/SFDO-Tooling/MetaCI/pull/411/commits/77cb7d05190fc5ee7f486d015d278197827ea223

src/js/components/error.js
prescod: 👍

PR #414, opened 2019-04-18
Docs and defaults to simplify setup.

prescod:


CONTRIBUTING.rst
```diff
@@ -31,11 +31,13 @@ Install Python requirements::
...
-Copy the ``.env`` file somewhere that will be sourced when you need it::
+Copy the ``.env`` file to config/settings/.env::
 
-    cp env.example $VIRTUAL_ENV/bin/postactivate
+    cp env.example config/settings/.env
 
-Edit this file to fill in values for the missing settings.
+Edit this file to fill in values for the missing settings, especially
+GITHUB_USERNAME and GITHUB_PASSWORD. You will need to use a
+`Personal Access Token` instead of a password if you have 2-factor auth setup.
```
jstvz: Quibble: "...have 2-factor authentication set up." perhaps?

CONTRIBUTING.rst
```diff
@@ -81,20 +85,22 @@ and running locally::
...
 
     ./manage.py metaci_scheduled_jobs
 
 
-Copying the production database
--------------------------------
+Alternative: Copying the production database
+---------------------------------------------
+
+** PLEASE CONTACT THE METACI TEAM BEFORE DOING THIS TO DISCUSS THE SECURITY IMPLICATAIONS **
```
jstvz: Maybe a `WARNING: (...)` will suffice? Noting that production secrets are included in the copy and warning users to beware is more generally applicable to external users.

prescod: @davisagli We are leaning towards just taking this section out.

CONTRIBUTING.rst
davisagli: And I think you need an underscore after the backticks for it to count as a reStructuredText link.

CONTRIBUTING.rst
davisagli: Yeah let's take it out.

prescod: Done.

https://github.com/SFDO-Tooling/MetaCI/pull/414/commits/036996e0d0b1ab266421eeec73a8f575c200a6e7

CONTRIBUTING.rst
prescod: Done and ready to review.

PR #426, opened 2019-05-10
Batch writes to the database.

prescod:


metaci/testresults/models.py
```diff
@@ -541,11 +542,20 @@ def summarize_week(cls, date):
...
+            while item:
+                yield item
+                item = list(itertools.islice(it, size))
+
+        for batch in split_seq(method_contexts, 5000):
+            new_objects = [cls(**values) for values in batch]
+            cls.logger.info("Creating %s for %s", len(new_objects), date)
+            created = cls.objects.bulk_create(new_objects)
+            cls.logger.info("Created %s for %s", len(created), date)
+        return
```
davisagli: Can you take out the no-op `return` and move `split_seq` to metaci.utils?

prescod: Thanks for merging master. I made the changes you suggested.

PR #427, opened 2019-05-13
Flow types and config files

prescod:
Added config files for linting, printing and node versioning.

Updated yarn.lock file to have the latest dependencies required to run all of the new tools.

Added flow typings to get rid of flow errors relating to missing type signatures.


prescod:
> I must confess I don't really know how to evaluate this. I'm assuming the eslint configuration is based on a recommendation from somewhere and the flow-types are generated automatically?

The ESLint came from sfdo-template and yes the flow-types are generated by a process recommended by Oddbird. I had to clean up some errors that the generator introduced but the cleanup was also pretty "mechanical."

The gist is that it declares everything as type "any" and then you hope that sometime in the future somebody, somewhere in the world (or on your team) will get around to carefully writing "real" type declarations.

davisagli:
@prescod can we add a section about that process to CONTRIBUTING.rst?

PR #441, opened 2019-05-29
Add a way to skip release-phase django migrations using an environment variable

davisagli:
This paves the way to running worker dynos in a separate app so that we can deploy the web app without interrupting builds in progress, without worrying about accidentally running the migration in both apps concurrently.

The web app will be deployed more frequently and always run the migrations. The worker app will only be deployed when necessary and can assume the migrations were already run.

When deploying both apps at the same time, it's mostly okay if the worker app starts
before the migrations are complete because no builds will be triggered until the web app is being served. Although there's a small risk of scheduled jobs running too early. If that's unacceptable we'll need a more complex solution with a db-level lock.


prescod:
This mechanism will work when migrations are "backwards compatible" with the worker dynos but if we added (for example) a required column to a table that the dyno writes to, the dyno would crash on attempt-to-write.

This isn't a show-stopper, but it is a risk.

PR #443, opened 2019-06-04
Add lwc-jest to package.json

jstvz:
This PR adds `lwc-jest` to provide preliminary support for unit testing Lightning Web Components. 

- TODO: I expect that we'll need to update SFDO-Tooling/CumulusCI-Test to
include a `lwc-jest` test or two.

- TODO: Capture test results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "@salesforce-ux/design-system": "^2.8.3",
     "@salesforce/design-system-react": "^0.9.6",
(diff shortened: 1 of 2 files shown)


prescod:
Will there be a new task for invoking this?

jstvz:
> Will there be a new task for invoking this?

I'd think so, possibly, but I haven't discussed with @davisagli. The intent of starting with MetaCI is to unblock a custom task in another project.

PR #444, opened 2019-06-07
For Review Purposes Only

jgerigmeyer:
porpoise


src/js/store/perfdata/selectors.js
```diff
@@ -0,0 +1,44 @@
...
+} from 'api/testmethod_perf_UI_JSON_schema';
+
+export const selectPerfState = (appState: AppState): PerfDataState =>
+  appState.perfData;
+
+export const selectPerfUIStatus = (appState: AppState): LoadingStatus =>
+  appState.perfDataUI ? appState.perfDataUI.status : 'LOADING';
+
+export const selectPerfDataAPIUrl = (appState: AppState): string =>
+  appState.perfData.status === 'AVAILABLE' ? appState.perfData.url : '';
```
jgerigmeyer: When selecting nested state, I would use reselect's `createSelector` helper to allow one selector to rely on another. See https://github.com/SFDO-Tooling/MetaDeploy/blob/master/src/js/store/products/selectors.js#L58-L72 for a simple example.

src/js/components/perfPages/perfPage.js
```diff
@@ -0,0 +1,176 @@
...
+          queryparams={queryparams}
+          items={results}
+        />
+      </div>{' '}
+      <DebugIcon />
+    </div>
+  );
+};
+
+const AuthError = ({ message }: { message: string }) => (
```
jgerigmeyer: This component isn't used (or even exported), causing a lint error.

src/js/components/perfPages/formFields/index.js
```diff
@@ -0,0 +1,115 @@
...
+  filterDef: FilterDefinition,
+  currentValue?: string | null,
+  onSubmit: () => typeof undefined,
+) => {
+  const fieldType = FieldTypes[filterDef.field_type];
+  if (fieldType) {
+    return fieldType(filterDef, currentValue, onSubmit);
+  }
+  // eslint-disable-next-line no-console
+  console.log('Unknown filterDef type', filterDef.field_type, filterDef);
```
jgerigmeyer: Use `log` helper from `utils/logging.js`?

src/js/components/perfPages/perfDataTable.js
```diff
@@ -0,0 +1,174 @@
...
+  queryparams,
+}: Props) => {
+  /*
+   * Extract the page from the server-generated URL and ensure it is
+   * browser URL before fetching it.
+   */
+  const goPageFromUrl = (url: string) => {
+    const qs = url.split('?', 2)[1];
+    const qParts = queryString.parse(qs);
+    const page = qParts.page;
```
jgerigmeyer: Could use `getUrlParam` helper from `utils/api.js`?

src/js/components/perfPages/perfTableUtils.js
```diff
@@ -0,0 +1,60 @@
...
+
+// This differs from the normal query paramemter helper classes in that
+// it is designed to have default parameters as state. Also, its fairly
+// strongly typed.
+export class QueryParamHelpers {
+  default_params: { [string]: mixed };
+
+  constructor(default_params: { [string]: mixed }) {
+    this.default_params = default_params;
+    this.get = this.get.bind(this); // Javascript grossness!
```
jgerigmeyer: Hm, you're using the public class fields syntax (that is, an arrow fn to define `get`), so I don't think this line is needed.

https://reactjs.org/docs/handling-events.html

PR #447, opened 2019-06-13
Feature/open permissions

prescod:
I forgot that Django's definition of "staff" does not mean "employee" but is closer to "admin". So many people we would consider "staff" can't use it. The first commit fixes that.

I've decided to increase the page timeout. So the second commit does that.

I noticed that if I want to delete a PlanSchedule it would delete its associated builds too due to a ForeignKey DELETE rule. The third commit fixes that.


metaci/api/views/testmethod_perf.py
```diff
@@ -323,7 +323,7 @@ def _check_params(self):
 
     def get_queryset(self):
         """The main method that the Django infrastructure invokes."""
-        set_timeout(10)
+        set_timeout(60)
```
davisagli: Heroku will time out after 30 seconds https://devcenter.heroku.com/articles/request-timeout

metaci/build/models.py
```diff
@@ -127,7 +127,7 @@ class Build(models.Model):
     org_instance = models.ForeignKey('cumulusci.ScratchOrgInstance',
                                      related_name='builds', null=True, blank=True, on_delete=models.CASCADE)
     schedule = models.ForeignKey('plan.PlanSchedule', related_name='builds',
-                                 null=True, blank=True, on_delete=models.CASCADE)
+                                 null=True, blank=True, on_delete=models.SET_NULL)
```
davisagli: You'll need to generate a migration for this.

Maybe we should have a test that runs a makemigrations dry run and breaks if it did anything.

src/testperf.html
```diff
@@ -8,7 +8,7 @@
     -->
   </head>
   <body>
-      {% if request.user.is_staff %}
+      {% if request.user %}
```
davisagli: Should we put the same condition around the tab that links here?

metaci/build/models.py
prescod: Yes it does generate a migration but based on the Django documentation it is my belief that the migration doesn't really do anything. It will end up re-creating the same ForeignKey with the same SQL but different Python behaviour. I guess I'll create it now to avoid some future programmer confusion about where it came from...

Can you remind me though: will deploying to prod run the migration automatically?

src/testperf.html
prescod: I think the UI is better this way given how often people forget to login. Instead of "I could have sworn such and such a feature exists, where is it?" I prefer "Oh...the site tells me I'm not logged in."

metaci/build/models.py
prescod: 3a6522f29309aac7f53d6a5085e62bff6769b711

metaci/api/views/testmethod_perf.py
prescod: 83ffecae2f9cbda99a7c54c9b6be9742d78968dc

metaci/build/models.py
davisagli: I don't think the migration will actually change anything in the database, but it's still needed for the internal bookkeeping of the migration system. If the behavior of a subsequent migration would be affected by this change in the field setting, the migration system needs to know that this change happened first, I guess.

Yes, deploying will run the migration. That's what the `release` line in Procfile is for.

src/testperf.html
davisagli: Oh right, I was thinking this is a situation where the user just sees a raw 403, but since it's a condition within the page it's better than that.

prescod: I fixed it and as penance for my mistake I also added tests.

davisagli: Looks like the test is failing here -- maybe some difference between template configuration between test/local/production settings?

prescod: No good deed goes unpunished. 

I'd better disable the test now and add a ticket to look into this later. Its some interaction between build and test phases. The permission issue is blocking users and getting it enabled for our internal users is actually higher priority than keeping it secret from outsiders. The security is more of a nice to have to discourage DDOS attacks through slow queries. And I'm 95% sure it works now based on my desk-testing, whether the unit tests pass or not.

prescod: Or to be more clear: the tests *do* pass on my laptop and my desk-checking passes but the tests don't run on CI due to a build/test interaction issue that I don't have time to investigate today.

davisagli: I'll merge this so we can get it deployed. I suspect the issue is that webpack is not running in the test setup, and that's how testperf.html gets generated.

PR #450, opened 2019-06-14
Various small metaci perf viewer cleanups

prescod:
3 unrelated commits.

One is a hack to clean up the UI and keep some related fields together on the same line. There are better ways but this UI will be replaced in the not too distant future and the better ways would involve more code to delete later.

The second deletes the last references to a feature that was obsoleted a month ago.

The last cleans up the URL handling to ensure that links can be passed around reliably.

### DIFF
--- a/src/js/components/perfPages/perfPage.js
+++ b/src/js/components/perfPages/perfPage.js
@@ -71,10 +71,12 @@ export const UnwrappedPerfPage = ({
      * path into query params with other filters
      */
     const pathParts = window.location.pathname.split('/');
-    const repo = pathParts[pathParts.length - 2];
-    const branch = 'master';
+    const default_repo = pathParts[pathParts.length - 2];
+    const default_branch = 'master';
+    const repo = (queryparams.get('repo') || default_repo).toString();
+    const branch = (queryparams.get('branch') || default_branch).toString();
     queryparams.set({ repo, branch });
-    doPerfRESTFetch({ ...queryparams.getAll(), repo, branch });
+    doPerfRESTFetch({ repo, branch, ...queryparams.getAll() });
     doPerfREST_UI_Fetch();
   }, []);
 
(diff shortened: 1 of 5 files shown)


src/js/components/perfPages/perfPage.js
```diff
@@ -71,10 +71,12 @@ export const UnwrappedPerfPage = ({
...
     const pathParts = window.location.pathname.split('/');
-    const repo = pathParts[pathParts.length - 2];
-    const branch = 'master';
+    const default_repo = pathParts[pathParts.length - 2];
+    const default_branch = 'master';
+    const repo = (queryparams.get('repo') || default_repo).toString();
+    const branch = (queryparams.get('branch') || default_branch).toString();
     queryparams.set({ repo, branch });
-    doPerfRESTFetch({ ...queryparams.getAll(), repo, branch });
+    doPerfRESTFetch({ repo, branch, ...queryparams.getAll() });
```
davisagli: Not quite following this. In what case is the `toString()` needed? Why do repo & branch need to be passed separately when they've just been set in queryparams?

prescod: The toString is to make the type checker happy because get() could return an integer, in theory.

On your second question: I was just adjusting the code that was there as opposed to re-thinking it carefully. You're right that it has probably been redundant for the last few revisions. I'll fix that.

prescod: So while I was looking through the code more deeply to ensure I hadn't missed anything, I realized that there is actually a mechanism that I wrote to allow the defaults to be set on the server side which is preferable because all of the knowledge of the data model is centralized there. Basically, I had anticipated this requirement but forgot that I had. Now the code takes advantage of that mechanism and also of your observation that I don't need to set "repo" in two different places.

Repo is still handled "specially" because it comes out of the URL. It's not a static default like the others. It could probably also be handled (dynamically) server-side but that's outside of the scope for today.

PR #457, opened 2019-07-02
Planrepos autopopulate

prescod:
I noticed that scheduled planrepos are not showing up in my review UI.

This turns out to be because they are not getting PlanRepos.

This turns out to be because the old PlanSchedule code that I revived doesn't set the PlanRepo.

This is probably because PlanSchedules predate PlanRepos.

Rather than just set the PlanRepo in the PlanSchedule code, I thought it would be better to fix it at a lower level so that if anyone forgot in any context in the future, the \_\_init\_\_() or save() would infer them.

Of course there were no tests for Build \_\_init\_\_ or save() so I had to add a few of those too...

### DIFF
--- a/metaci/build/test_models.py
+++ b/metaci/build/test_models.py
@@ -12,8 +12,7 @@ from metaci.conftest import (
 # Create your tests here.
 class BuildTestCase(TestCase):
     def test_empty_build_init(self):
-        build = Build()
-        self.assertTrue(build)
+        Build()
 
     def test_scheduled_build_init(self):
         repo = RepositoryFactory()
--- a/metaci/build/test_models.py
+++ b/metaci/build/test_models.py
@@ -20,6 +20,9 @@ class BuildTestCase(TestCase):
         branch = BranchFactory(name="branch", repo=repo)
         schedule = PlanScheduleFactory(branch=branch)
         plan = PlanFactory()
+
+        planrepo = PlanRepositoryFactory(plan=plan, repo=repo)
+
         build = Build(
             repo=branch.repo,
             plan=plan,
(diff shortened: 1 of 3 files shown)


prescod:
I guess I misunderstood the side effects of creating a PlanRepository.

I'm still confused. PlanRepositories don't refer to branches, so how could it trigger builds? Don't you need a branch or some kind of branch specification to trigger build?

I have been thinking of PlanRepositories as primarily a hook for hanging permissions upon.

Wouldn't it be a "Plan Repository Trigger" which triggers builds?

davisagli:
PlanRepository is what connects a plan to particular repositories. When a github webhook comes in, if there is an active planrepo for the repository where the plan's trigger == 'commit' and plan's regex matches the branch from the webhook, that will create a new build.

PlanRepositoryTrigger is for a different use case: triggering a build in repo Y from builds in repo X (i.e. Abacus depends on NPSP and wants to run a build whenever there are changes to its dependency)

metaci/build/test_models.py
```diff
@@ -0,0 +1,43 @@
...
+    PlanRepositoryFactory,
+    PlanScheduleFactory,
+)
+
+
+# Create your tests here.
+class BuildTestCase(TestCase):
+    def test_empty_build_init(self):
+        build = Build()
+        self.assertTrue(build)
```
davisagli: What's the value of this test?

prescod: The virtue of having the test at all is to ensure that my \_\_init\_\_ code works when no arguments are passed.

The virtue of asserting that it didn't return null is very minor so I removed it in ceb06d4f34dd7dac72d91cee8108ffb7e91eaf96

PR #466, opened 2019-08-09
Make it possible to store Robot Framework screenshots as non-public S3 objects

davisagli:
This changes the code which imports robot test result XML to convert screenshot URLs into an intermediate representation like asset://1234 (using the id of the TestResultAsset in the database) instead of directly into a public S3 URL. And it changes the view which renders those results to convert the intermediate URL into an S3 URL that includes time-limited query string authorization.

This is meant for use with the new S3 buckets we're moving to. Existing test results still keep their old screenshot URLs, which will stop working once the old S3 buckets are deleted.

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -86,7 +86,9 @@ AWS_ACCESS_KEY_ID = env("DJANGO_AWS_ACCESS_KEY_ID")
 AWS_SECRET_ACCESS_KEY = env("DJANGO_AWS_SECRET_ACCESS_KEY")
 AWS_STORAGE_BUCKET_NAME = env("DJANGO_AWS_STORAGE_BUCKET_NAME")
 AWS_AUTO_CREATE_BUCKET = True
-AWS_QUERYSTRING_AUTH = False
+AWS_BUCKET_ACL = "private"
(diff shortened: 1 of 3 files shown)


davisagli:
@jlantz here's a test result on staging using the new S3 links. https://mrbelvedereci-staging.herokuapp.com/tests/result/598409

boakley:
Have you explicitly tested this with screenshots taken in a suite or test teardown? It was mentioned in Safiya's _Robot/MetaCI issues & solutions_ document that those screenshots are (sometimes? always?) being lost in the current implementation. If you're mucking about in the code that preserves the screenshots, this is an edge case that probably ought to be specifically tested for.

davisagli:
@boakley I haven't. Do you know where I can find an example of a test run where those screenshots were lost?

boakley:
@davisagli  here is a build that failed during setup and where the screenshots don't seem to have been preserved:

https://mrbelvedereci.herokuapp.com/tests/result/55509622

davisagli:
@boakley Thanks. Confirmed and I can see in the code for importing robot test results that we're only preserving screenshots within test cases, not in setup/teardown. I opened an issue #469

PR #468, opened 2019-08-13
Make it possible to configure multiple comma-separated github webhook secrets

davisagli:
This way we can rotate the secret with zero downtime.


davisagli:
Actually on second thought I don't think we need this. I'll set up the webhook in the github app with a new secret, so MetaCI will ignore it until I update the GITHUB_WEBHOOK_SECRET environment variable to the new secret, and then it will ignore the old webhooks.

Issue #464, opened 2019-08-02
Robot import errors aren't being reported


boakley:
If a robot test imports a library, and that library can't be imported for some reason (not installed, or has a syntax error), robot displays an error on stdout. In addition, the error is mentioned in the robot log before any of the test results, in a section titled "Test Execution Errors". This section appears in the log before the section titled "Test Execution Log".

This sort of error is not available in MetaCI, and it's critical for being able to debug failures that happen in MetaCI due to import errors rather than test failures inside a specific test. 

We need to modify MetaCi to detect any of these "pre-test" errors and display them.

boakley:
I've attached an output.xml file which shows one of these types of errors. It was generated by this simple test case:

```
*** Settings ***
Library  bogus

*** Test Cases ***
Whatever
    log  whatever!
```

(I had to give the file a .txt suffix; github wouldn't let me attach .xml oddly enough)

[output.xml.txt](https://github.com/SFDO-Tooling/MetaCI/files/3602771/output.xml.txt)

PR #479, opened 2019-09-11
Import Robot setup/teardown screenshots

### LINKED ISSUE #469: Screenshots from robot setup/teardown aren't transferred to S3
Currently the logic for persisting screenshots during import of robot results is only being done for test cases and not for setup/teardown.

davisagli:
This adjusts the existing code for importing Robot Framework results to handle screenshots from setup/teardown. (Fixes #469)

These are stored as a BuildFlowAsset since they aren't related to a specific TestResult.

Unrelatedly I also adjusted the populate_db script to default to setting up the CumulusCI-Test repository (the old script for adding fake builds and test results is still available via the --results option)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -204,39 +204,3 @@ Developing with SLDS
 MetaCI uses https://github.com/SalesforceFoundation/django-slds which imports version 2.1.2 of the Salesforce Lightning Design System.
 
 You can find a CSS and component reference archived here: https://archive-2_1_2.lightningdesignsystem.com/
-
-Setting up MetaCI with CumulusCI-Test
--------------------------------------
-
-To connect your local MetaCI to the CumulustCI-Test repository you will need to perform the following:
-- Login to the Django Admin page (`localhost:8000/admin/login`)
-- Scroll down and click on the "Repositories" object
-- In the top right click the "Add Repository +" button
-- Enter the following:
-    - Name: CumulusCI-Test
-    - Owner: SFDO-Tooling
-    - Url: https://github.com/SFDO-Tooling/CumulusCI-Test
-- Click 'Save'
-- Note: You'll want to ensure that you have the github variables set in your corresponding env.example file.
-
-In order to run a build against the CumulusCI-Test repository you will also need to create an `org` to associate with the new CumulusCI-Test repository.
-
-- Navigate to the Django Admin page
-- Under the 'CumulusCI' section click on 'Orgs'
-- In the top right click the "Add Org +" button
-- Enter the following to setup a dev org:
-    - Name: dev
-    - Json: {"config_file":"orgs/dev.json","scratch":true}
-    - Scratch (checkbox): checked/true
-    - Repo (picklis): SFDO-Tooling/CumulusCI-Test
-- Click the 'SAVE' button
-
-Finally, you'll want to ensure that the "Dev Org" plan is associated with the CumulusCI-Repository:
-
-- Navigate to the Django Admin page
-- Under 'Plan' click on 'Plans'
-- Click on 'Dev Org'
-- Scroll down to the bottom to the 'Plan Repositories' section:
-    - Click 'Add another Plan repository
-    - Select the 'SFDO-Tooling/CumuluCI-Test repository
-    - Click the 'SAVE' button
--- a/metaci/db/management/commands/populate_db.py
+++ b/metaci/db/management/commands/populate_db.py
@@ -1,59 +1,71 @@
-from django.core.management.base import BaseCommand
-
-from django.db import transaction
-
-from guardian.shortcuts import assign_perm
+import random
 
 import factory.random
+from django.contrib.auth.models import Group
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from guardian.shortcuts import assign_perm
 
 from metaci import conftest as fact
-from metaci.users.models import User
-from metaci.testresults.models import TestResult
-from django.contrib.auth.models import Group
-
 from metaci.build.models import BuildFlow
-
-import random
+from metaci.cumulusci.models import Org
+from metaci.plan.models import Plan, PlanRepository
+from metaci.repository.models import Repository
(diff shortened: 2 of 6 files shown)


metaci/db/management/commands/populate_db.py
```diff
@@ -1,59 +1,71 @@
...
+            "--results", action="store_true", help="Create fake builds and test results"
+        )
 
-            self.create_objs()
-        else:
-            print(
+    def handle(self, *args, **options):
+        if options["results"]:
+            if self.check_non_destructive():
+                print("Populating test data")
```
boakley: I'm not familiar with this code -- are print statements how MetaCI does logging? Or were these some debugging statements that were accidentally left in?

metaci/db/management/commands/populate_db.py
```diff
@@ -1,59 +1,71 @@
...
+                # reset_db.Command().handle(*args, **options)
```
boakley: Is there a reason these were left in but commented out?

metaci/db/management/commands/populate_db.py
```diff
@@ -1,59 +1,71 @@
...
+                # migrate.Command().handle(*args, **options)
+
+                self.create_builds_with_test_results()
+            else:
+                print(
+                    """
+                This does not seem to be an empty or test database!
```
boakley: Salesforce writing guides recommend not using exclamation points in error messages and instructional text. I recommend removing the exclamation mark on this and the next line. 

The [writing style guidelines trail ](https://trailhead.salesforce.com/content/learn/modules/writing_style/writing_style_guidelines)says this:

> - Use exclamation points only to be encouraging or to generate excitement.
> - Don’t use exclamation points in error messages, confirmation messages, or instructional text.

metaci/db/management/commands/populate_db.py
davisagli: All these specific lines you commented on are code that Paul wrote a while back and I didn't touch aside from moving inside a conditional.

Using a logger would be best but this code is in a command line command so the print function isn't doing any harm.

metaci/db/management/commands/populate_db.py
davisagli: I assume because it's an occasionally useful debugging aid.

metaci/db/management/commands/populate_db.py
davisagli: Good to know, though so far the only real audience for this command is Paul.

CONTRIBUTING.rst
```diff
@@ -204,39 +204,3 @@ Developing with SLDS
 MetaCI uses https://github.com/SalesforceFoundation/django-slds which imports version 2.1.2 of the Salesforce Lightning Design System.
 
 You can find a CSS and component reference archived here: https://archive-2_1_2.lightningdesignsystem.com/
-
```
prescod: What's the story behind this deletion?

metaci/db/management/commands/populate_db.py
boakley: fair enough.

prescod: Why did you prefer this part of the data generation be optional?

metaci/db/management/commands/populate_db.py
```diff
@@ -1,59 +1,71 @@
...
         return User.objects.count() < 3 and TestResult.objects.count() < 10
 
-    def create_objs(self):
+    def create_builds_with_test_results(self):
         with transaction.atomic():
             factory.random.reseed_random("TOtaLLY RaNdOM")
             random.seed("RaNDOM!! TOtaLLY")
             fact.UserFactory()
             fact.UserFactory()
+            existing_plans = Plan.objects.all()
```
prescod: Why do you rely on plans existing instead of creating them? If this is a fresh database where would they come from? Have I missed something?

metaci/testresults/importer.py
```diff
@@ -163,6 +164,24 @@ def import_robot_test_results(build_flow, path):
...
+        for screenshot in result["suite"]["screenshots"]:
+            if screenshot in suite_screenshots:
+                continue
+            screenshot_path = screenshot
+            if dirname:
+                screenshot_path = "{}/{}".format(dirname, screenshot)
+            with open(screenshot_path, "rb") as f:
+                asset = BuildFlowAsset(
+                    build_flow=build_flow,
+                    asset=ContentFile(f.read(), screenshot),
```
prescod: This isn't a problem with this PR but I'm surprised that we store photos in the DB instead of on S3.

CONTRIBUTING.rst
davisagli: @prescod my changes to the populate_db script (mentioned above in this doc) now take care of doing these steps automatically.

metaci/db/management/commands/populate_db.py
davisagli: @prescod there are some default plans created in one of the database migration steps. The PlanRepositoryFactory was relying on those before too, I just moved that here so that the PlanRepositoryFactory is also useful in situations where there are no existing plans, such as a test case.

metaci/testresults/importer.py
davisagli: @prescod that's not the case. ContentFile is a thing from django-storages which takes care of sending the file to S3 and holding a reference to it.

boakley:
One of the other problems we've had in the past is that errors that happen at startup get lost, such as libraries failing to load. By any chance is that problem at all related to this problem? I don't want to introduce scope creep, but if you're mucking about in the code...

See #464

davisagli:
@boakley let's handle that separately. It would help if you could show me a "normal" (i.e. non-MetaCI-mutilated) output.xml for a test run that includes the import errors, so I can see where they end up.

boakley:
> @boakley let's handle that separately. It would help if you could show me a "normal" (i.e. non-MetaCI-mutilated) output.xml for a test run that includes the import errors, so I can see where they end up.

Sounds good. I've attached an example output.xml file to #464 that includes an import error.

PR #486, opened 2019-09-24
Default date range on page load

prescod:
testmethod_perf.py changed to add a default date range so that we wouldn't try to provide results to the beginning of time by default.

perfPage.js changed to fetch the UI properties and defaults from the server before fetching the data. This is very slightly slower, but it manages an architectural convention that the client doesn't know much about the details of the schema, data model or defaults.

### DIFF
--- a/metaci/api/views/testmethod_perf.py
+++ b/metaci/api/views/testmethod_perf.py
@@ -14,12 +14,16 @@ from metaci.repository.models import Repository, Branch
 from metaci.plan.models import Plan
 
 from django.db import connection
+from datetime import datetime, timedelta
+
+thirty_days_ago = datetime.now() - timedelta(days=30)
 
 
 class DEFAULTS:
     page_size = 50
     max_page_size = 100
     branch = "master"
+    daterange_after = thirty_days_ago.isoformat("T").split("T")[0]
 
 
 def set_timeout(timeout):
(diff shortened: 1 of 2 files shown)


metaci/api/views/testmethod_perf.py
```diff
@@ -14,12 +14,16 @@
 from metaci.plan.models import Plan
 
 from django.db import connection
+from datetime import datetime, timedelta
+
+thirty_days_ago = datetime.now() - timedelta(days=30)
```
davisagli: I was gonna object to calculating this at import time but I guess it'll be okay on Heroku with the dynos restarting daily.

src/js/components/perfPages/perfPage.js
```diff
@@ -65,18 +65,24 @@ export const UnwrappedPerfPage = ({
...
-    doPerfRESTFetch({ ...queryparams.getAll() });
     doPerfREST_UI_Fetch();
   }, []);
 
+  // Fetch the real data
+  useEffect(() => {
+    if (uiAvailable) {
+      doPerfRESTFetch({ ...queryparams.getAll() });
+    }
+  }, [uiAvailable]);
```
davisagli: This forced me to finally read about React Hooks a bit. Cool.

PR #489, opened 2019-09-25
New build/worker scheduler

davisagli:
Key improvements:
- Plans can be prioritized. When there's a backlog, higher priority builds will be chosen first.
- MetaCI can autoscale its own worker pool

This is still in progress and needs more testing. Before working on unit tests I'd like to deploy this to staging (without merging) and see what issues I uncover there.


davisagli:
Closing in favor of #499 which achieves a similar end result with less invasive changes to the codebase.

PR #501, opened 2019-10-03
Jest Tests

prescod:
Wait for other PRs to be merged and then do this one.


src/test/js/.eslintrc.yml
```diff
@@ -0,0 +1,16 @@
...
+  sourceType: module
+env:
+  browser: true
+  commonjs: true
+  jest: true
+  node: false
+rules:
+  import/namespace:
+    - 2
+    - allowComputed: true
```
davisagli: Is the point of this to disable rules from the top-level .eslintrc.yml, or what?

prescod: Yes it overrides the rules from the top-level file. It came from the template project:

https://github.com/SFDO-Tooling/sfdo-template/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/test/js/.eslintrc.yml

The main purpose is presumably to make eslint "jest-aware". Without it you get:

   94:5  error  'expect' is not defined    no-undef

I haven't researched what each specific line does, but the file as a whole is needed.

prescod:
I'll open a new PR for heroku_ci.sh because its logically a separate feature and I want to merge all Javascript-heavy branches.

PR #503, opened 2019-10-03
Accumulate exit status and combine coverage testing

prescod:


heroku_ci.sh
```diff
@@ -5,9 +5,17 @@ git clone -b "$HEROKU_TEST_RUN_BRANCH" --single-branch https://github.com/SFDO-T
 cd MetaCI_checkout
 git reset --hard $HEROKU_TEST_RUN_COMMIT_VERSION
 export DJANGO_SETTINGS_MODULE=config.settings.test
-coverage run $(which pytest) --tap-stream
```
davisagli: We still need this command to actually run the tests (and incorporate its exit status).

davisagli: Unless I'm jumping the gun and you're just removing it temporarily to iterate on the other stuff...

prescod: Yeah I've been experimenting with using draft pull requests as a way of jumping between work items, but I guess you tend to review any PRs you see, whether I tag you or not. I don't strictly need PRs for everything but they are harder to lose track of than a branch.

My actual challenge with this now that I've iterated a bit is that I don't know where to get the build number.

The Heroku equivalent of CIRCLE_BUILD_NUM would seem to be HEROKU_RUN_ID but Coveralls doesn't like it.

I couldn't find any documentation for the github/heroku ci/coveralls combination. 

https://docs.coveralls.io/supported-ci-services

prescod: I have a lot of gaps in my knowledge of the three services (plus CircleCI, which I am trying to "port" from) so either you can finish this up or you can give me a brain-dump tomorrow.

davisagli: I think that's a fine use of draft pull requests; I just didn't see that it was draft until after I wrote my first comment.

I also couldn't find documentation but looking at the code for the Python coveralls client, it looks like we can trick it into using the CircleCI vars if we do this prior to invoking coveralls:

    export CIRCLECI=true
    export CIRCLECI_BUILD_NUM=$HEROKU_RUN_ID

davisagli:
@prescod all right, this seems to work

PR #508, opened 2019-10-10
First half of build-stopping feature

prescod:
New abstraction called "WorkEnvironment" which abstracts away the Heroku API, Heroku environment variable(s) and some details of working with Heroku.

Uses redis cache to build a mapping from build numbers to Dynos.

Adds a stop method but not a UI for it.

Not tested at all: I'd need to do pytests and tests on Staging.


metaci/build/workenvironment.py
```diff
@@ -0,0 +1,78 @@
...
+    def scale_to(self, quantity):
+        pass
+
+    @abstractmethod
+    def stop_worker(self, worker_id):
+        pass
+
+    @abstractmethod
+    def get_worker_id_from_environment(self):
+        pass
```
davisagli: docstrings please

metaci/build/workenvironment.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def scale_to(self, quantity):
+        url = f"https://api.heroku.com/apps/{settings.HEROKU_APP_NAME}/formation/worker"
+        resp = requests.patch(url, json={"quantity": quantity}, headers=self.headers)
+        resp.raise_for_status()
+
+    def stop_worker(self, worker_id):
+        url = (
+            f"https://api.heroku.com/apps/{settings.HEROKU_APP_NAME}/dynos/{worker_id}"
+        )
```
davisagli: Let's add a _url method to build a Heroku Platform API URL so we don't have to repeat the base portion.

prescod:
So what direction do you want this to go in, given that I can't commit to doing all of the authorization etc. needed to do a proper UI?

1. I could roll back to a 2-line patch that just puts the Dyno ID in the build log.

2. I could make a single class abstraction for all Heroku "stuff".

3. I could add a "stop" method to the abstraction.

4. I could add a key to Redis so that any instance can look up the InstanceID to call a Stop method.

That's about the limit of what I can commit to in October. Which of those would you like me to do?

davisagli:
> I could roll back to a 2-line patch that just puts the Dyno ID in the build log.

@prescod Just doing this sounds like by far the lowest hanging fruit. I'm pretty in favor of point number 2 as well. I'd rather wait on 3 and 4 until I can play around and make sure we have a good way to avoid the problems I mentioned above.

PR #510, opened 2019-10-11
Infrastructure for Typescript alongside Flow (for now)

prescod:
I think these changes are non-breaking and independent of everything else, so might as well split them out.

Adds Typescript handling to config files and also removes the unused and obsolete Raven.


package.json
```diff
@@ -54,14 +67,15 @@
...
-    "eslint-plugin-flowtype": "^4.3.0",
-    "eslint-plugin-flowtype-errors": "^4.1.0",
-    "eslint-plugin-import": "^2.18.2",
-    "eslint-plugin-prettier": "^3.1.1",
-    "eslint-plugin-react": "^7.16.0",
+    "eslint-plugin-import": "^2.18.0",
+    "eslint-plugin-prettier": "^3.1.0",
+    "eslint-plugin-react": "^7.14.2",
+    "eslint-plugin-react-hooks": "^1.6.1",
+    "eslint-plugin-simple-import-sort": "^4.0.0",
```
davisagli: Why are we downgrading some of the dependencies?

PR #511, opened 2019-10-12
Remove user handling because we don't use it.

prescod:
Django uses cookies for user handling. Login is handled through HTML/HTTP, not Javascript/HTTP. These user actions never ended up being used in MetaCI.


prescod:
I think it *is* gone:

https://github.com/SFDO-Tooling/MetaCI/tree/93241c301cb768dd2dcb00dd567294586cda23ab/src/js/store

davisagli:
@prescod ah sorry, I was probably looking at a single commit thinking I was looking at the full diff for the PR.

PR #521, opened 2019-10-21
make worker dyno proctype configurable

cdcarter:
really, this just makes the intent of the code more understandable, i doubt anyone needs to toggle this but why not

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -255,8 +255,8 @@ if HIREFIRE_TOKEN:
 
 HEROKU_TOKEN = env("HEROKU_TOKEN", default=None)
 HEROKU_APP_NAME = env("HEROKU_APP_NAME", default=None)
-WORKER_DYNO_NAME = env("WORKER_DYNO_NAME", default="worker")
-if HEROKU_TOKEN and HEROKU_APP_NAME and WORKER_DYNO_NAME:
+WORKER_DYNO_NAME = env("WORKER_DYNO_NAME", default=None) or "worker"
+if HEROKU_TOKEN and HEROKU_APP_NAME:
     METACI_WORKER_AUTOSCALER = "metaci.build.autoscaling.HerokuAutoscaler"
 
 # Custom Admin URL, use {% url 'admin:index' %}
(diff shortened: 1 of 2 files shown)


metaci/build/autoscaling.py
```diff
@@ -95,9 +95,10 @@ def scale(self):
 
 class HerokuAutoscaler(Autoscaler):
     """Scale using Heroku worker dynos."""
+    API_ROOT = "https://api.heroku.com/apps"
 
     def scale(self):
-        url = f"https://api.heroku.com/apps/{settings.HEROKU_APP_NAME}/formation/worker"
+        url = f"{API_ROOT}/{settings.HEROKU_APP_NAME}/formation/{settings.WORKER_DYNO_NAME}"
```
davisagli: I think this will result in `NameError: name 'API_ROOT' is not defined`; class members aren't included in method scope in Python.

config/settings/production.py
```diff
@@ -255,7 +255,8 @@
 
 HEROKU_TOKEN = env("HEROKU_TOKEN", default=None)
 HEROKU_APP_NAME = env("HEROKU_APP_NAME", default=None)
-if HEROKU_TOKEN and HEROKU_APP_NAME:
+WORKER_DYNO_NAME = env("WORKER_DYNO_NAME", default="worker")
+if HEROKU_TOKEN and HEROKU_APP_NAME and WORKER_DYNO_NAME:
```
davisagli: How about `WORKER_DYNO_NAME = env("WORKER_DYNO_NAME", default=None) or "worker"` and then it doesn't need to be included in the conditional.

cdcarter: I guess it could just have the default and not be part of the condition, does default=None give us anything over default="worker" ?

metaci/build/autoscaling.py
cdcarter: heh yea you right

config/settings/production.py
davisagli: It gives us a better default if someone sets the env var to an empty string, but that's pretty edge case-y.

cdcarter: I did that, but i agree it's edge casey. Perhaps worth considering a deploy check that asserts the settings vals are meaningful and not just set. Throwing random nonsense in the env locally is still a useful trick but the prod environment should have well formed env. something to think about.

plus i just love django checks.

davisagli:
Thanks. I moved this to a local branch and opened #548

PR #525, opened 2019-10-24
Django Command: run builds from command-line.

prescod:
A django command that allows one to run builds from a command line with going through queues, auto-scalers and schedulers etc.

Further explanation is here:

https://salesforce.quip.com/UbmFApGBoIzQ


metaci/build/management/commands/run_build.py
```diff
@@ -0,0 +1,74 @@
...
+        else:
+            status = lock_org(build.org, build.pk)
+
+            assert status, "Could not lock org"
+
+        dyno = os.environ["DYNO"]
+        print(f"Running build {build.pk} in {dyno}")
+        with temp_disconnect_signal(
+            signal=signals.post_save, receiver=handlers.queue_build, sender=Build
+        ):
```
davisagli: I really don't like this. For one thing I'm not sure it's threadsafe. Instead my suggestion would be to add a new option for build_type ("one-off"?) and make the signal handler skip queueing the build if it's that type.

metaci/build/management/commands/tests/test_run_build.py
```diff
@@ -0,0 +1,42 @@
...
+    #     self._testrun_build(run_build, True)
+
+    def test_run_build_persistent(self, run_build):
+        self._testrun_build(run_build, False)
+
+    def test_run_build_persistent_2(self, run_build):
+        self._testrun_build(run_build, False)
+
+    def test_run_build_persistent_3(self, run_build):
+        self._testrun_build(run_build, False)
```
davisagli: What's the purpose of doing the same test 3 times?

metaci/build/management/commands/tests/test_run_build.py
```diff
@@ -0,0 +1,42 @@
...
+    def _testrun_build(self, run_build, org_is_scratch):
+        repo = RepositoryFactory(name="myrepo")
+        BranchFactory(name="mybranch", repo=repo)
+        plan = PlanFactory(name="myplan", org="myorg")
+        PlanRepositoryFactory(repo=repo, plan=plan)
+        UserFactory(username="username")
+        OrgFactory(name="myorg", repo=repo, scratch=org_is_scratch)
+        os.environ["DYNO"] = "worker.42"
+        c = Command()
+        c.handle("myrepo", "mybranch", "commit", "myplan", "username")
```
davisagli: You could assert that run_build was called, that the build it was called with has the expected values, and that nothing was added to the queue.

metaci/build/management/commands/run_build.py
prescod: I can change it, but I think your concern is more aesthetic than practical. Keep in mind that we are talking about the last line of a command line command.

davisagli: I guess you're right that it's not a practical problem in this context, but yeah, I don't want to add code that's temporarily messing with signal handlers, and I think my suggestion is still a pretty straightforward alternative that avoids it.

prescod: My preferred solution would be to get rid of the signal but I also don't want to take on that project right now.

davisagli: Yeah I agree with that as a longer term goal. Signals are for extensions/pluggability, not for assembling an application that we have full control over.

metaci/build/management/commands/tests/test_run_build.py
prescod: Done.

metaci/build/management/commands/run_build.py
prescod: Okay, it's changed now.

metaci/build/management/commands/tests/test_run_build.py
prescod: I was having a weird crashing behaviour when I ran more than one test. I got so fixated on fixing the crash that I forgot to put the tests back to what they were originally testing. In the end it was because I was trying to hook into the Command API at a low-level which would run through all of the argument validation code etc. but that low-level was also resetting Django state. It was a while between when I wrote the call to the low-level API and when it started causing trouble (because I originally had only a single unit test) so it wasn't immediately obvious to me that that was the source of the problem. Anyhow, I got distracted fixing that and forgot to put my tests back to their original state.

metaci/build/management/commands/tests/test_run_build.py
prescod: Done.

metaci/build/management/commands/tests/test_run_build.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.mark.django_db
+class TestRunBuild(TestCase):
+    # def test_run_build_scratch(self, run_build):
+    #     self._testrun_build(run_build, True)
+
+    def test_run_build_persistent(self, run_build, scratch_org_limits):
+        self._testrun_build(run_build, scratch_org_limits, False)
+
+    def test_run_build_persistent_(self, run_build, scratch_org_limits):
+        self._testrun_build(run_build, scratch_org_limits, True)
```
davisagli: Why does this test case have persistent in the name but sets org_is_scratch to True?

PR #527, opened 2019-10-25
Retry with smaller number if autoscaling limits are exceeded

prescod:
Detects how many to try with based on (max_workers - already_used_workers).

### DIFF
--- a/metaci/build/autoscaling.py
+++ b/metaci/build/autoscaling.py
@@ -97,7 +97,8 @@ class HerokuAutoscaler(Autoscaler):
     """Scale using Heroku worker dynos."""
 
     def scale(self):
-        url = f"https://api.heroku.com/apps/{settings.HEROKU_APP_NAME}/formation/worker"
+        worker_type = "worker"
+        url = f"https://api.heroku.com/apps/{settings.HEROKU_APP_NAME}/formation/{worker_type}"
         headers = {
             "Accept": "application/vnd.heroku+json; version=3",
             "Authorization": f"Bearer {settings.HEROKU_TOKEN}",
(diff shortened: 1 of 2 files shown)


metaci/build/autoscaling.py
```diff
@@ -114,8 +115,23 @@ def scale(self):
             resp = requests.patch(
                 url, json={"quantity": self.target_workers}, headers=headers
             )
+            if resp.json() and resp.json().get("id") == "cannot_update_above_limit":
+                limit = resp.json()["limit"]
+                resp = self.scale_max(url, headers, worker_type, limit)
+
             resp.raise_for_status()
```
davisagli: resp.raise_for_status() should happen before we try to parse the response as json.

metaci/build/autoscaling.py
```diff
@@ -114,8 +115,23 @@ def scale(self):
             resp = requests.patch(
                 url, json={"quantity": self.target_workers}, headers=headers
             )
+            if resp.json() and resp.json().get("id") == "cannot_update_above_limit":
```
davisagli: You tested scaling above the hard limit (100). I wonder if we get a different message if there are x dynos running and we try to scale above (100 - x)?

metaci/build/autoscaling.py
prescod: We don't want to throw the exception because we want to respond to the 422. Heroku error messages are JSON so it is safe to parse them.

But yeah, perhaps there is a bit more thought required here about if Heroku was literally down/unreachable or something and didn't return JSON: but the only improvement I can see would be around the clarity of the error message. Instead of "JSON failed to parse" we could return a bit more information about the response. And again, I think this is only in the case where there is a deep networking or infrastructure issue.

davisagli: Oh yeah I see. I think it's okay as is since the server formats errors as JSON too.

metaci/build/autoscaling.py
prescod: I did try that too. Updated to 99. Same message.

I'll double-check though.

PR #533, opened 2019-10-28
dependabot config file

Br4nd0R:
* New `.dependabot/config.yml` file


.dependabot/config.yml
```diff
@@ -0,0 +1,18 @@
...
+          dependency_name: 'pytest'
+          version_requirement: '4.x'
+      - match:
+          # Waiting for Bryan to review newest version
+          dependency_name: 'robotframework-seleniumlibrary'
+          version_requirement: '3.x'
+      - match:
+          # No published source tarbal for 0.74.3
+          dependency_name: 'simple-salesforce'
+          version_requirement: '0.74.3'
```
davisagli:
```suggestion
          version_requirement: '0.74.2'
```

0.74.3 is the one we're trying to avoid

PR #540, opened 2019-10-31
🛢 Add org_api_version to build.

prescod:
Clean PR without obsolete migrations etc.


metaci/build/models.py
```diff
@@ -325,6 +326,11 @@ def run(self):
             self.flush_log()
             return
 
+        try:
+            self.org_api_version = org_config.latest_api_version
+        except BaseException as e:
```
davisagli: I'll approve this, but Exception is probably a better choice than BaseException generally...if the user triggers a KeyboardInterrupt or something raises SystemExit we generally don't want to swallow that.

prescod: Okay I updated it and made another super-minor change.

PR #549, opened 2019-11-04
Join related repos when querying Branch

### LINKED ISSUE #542: When creating a notification, the list of branches includes every branch that ever had a build against it
When I try to add a notification from a specific branch at https://mrbelvedereci.herokuapp.com/notifications/ I'm seeing a list of branches that includes branches that were deleted years ago, such as feature/crlp-ocr-fix. There are hundreds more like that.

Is there a reason this page has to list every single branch that ever existed rather than jut the current branches that currently exist (or maybe branches that have had builds in the last 30 days)? This giant causes two issues.

1. The page times out on the first load
2. It's really challenging to find the actual branch I want to create a notification for

Implementing #28 would probably help with the 2nd point, but not the the first.

davisagli:
so we don't have to do a separate query for each branch to get the repo when displaying a list of branches in a dropdown

fixes #542

### DIFF
--- a/metaci/repository/models.py
+++ b/metaci/repository/models.py
@@ -6,6 +6,7 @@ from django.apps import apps
 from django.db import models
 from django.http import Http404
 from django.urls import reverse
+from model_utils.managers import SoftDeletableManager
 from model_utils.models import SoftDeletableModel
 
 from metaci.cumulusci.keychain import GitHubSettingsKeychain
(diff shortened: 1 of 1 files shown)


prescod:
Re #542: I understand how this code fixes the timeout but I don't understand how it fixes the fact that there are old branches showing up.

prescod:
I can approve it as an improvement, however.

davisagli:
@prescod thanks -- I didn't actually confirm whether old branches are included yet, since I couldn't load the page...this should make it easier to check.

PR #551, opened 2019-11-07
Refactor `github_push_webhook`

Br4nd0R:
I felt that the `github_push_webhook` method was a bit large and could be made more readable if it was refactored a bit.


davisagli:
@Br4nd0R would you be willing to work on some tests for it too?

Br4nd0R:
> @Br4nd0R would you be willing to work on some tests for it too?

Yes, definitely!

metaci/repository/views.py
```diff
@@ -210,4 +190,66 @@ def github_push_webhook(request):
...
+            branch.is_removed = False
+            branch.save()
+
+    return branch
+
+
+def get_branch_name_from_payload(push):
+    branch_ref = push.get("ref")
+    if not branch_ref:
+        return HttpResponse("No branch found")
```
Br4nd0R: @davisagli - Wondering what the best way to handle this case would be. We want this to be detected at the level of `github_push_webhook()` but this check is being performed a couple layers down from it. Not sure what the best way to go about this would be.

PR #607, opened 2019-11-11
Autoscaler for robot app

Br4nd0R:


config/settings/common.py
```diff
@@ -360,7 +360,8 @@ def url_prefix_list(val: str) -> List[str]:
 WORKER_DYNO_NAME = env("WORKER_DYNO_NAME", default=None) or "worker"
 
 # Queues that the given app will has
-HEROKU_APP_QUEUES = env("HEROKU_APP_QUEUES", default=("high", "medium", "default"))
+queues = env.list("HEROKU_APP_QUEUES", default=["high", "medium", "default"])
+HEROKU_APP_QUEUES = set(queues)
```
davisagli: This should be a list; not a set. There's a line in the autoscaling code which assumes all workers are processing the first queue, and turning this into a set loses the ordering.

metaci/build/autoscaling.py
```diff
@@ -99,29 +100,39 @@ class HerokuAutoscaler(Autoscaler):
...
-        url = f"{self.API_ROOT}/{settings.HEROKU_APP_NAME}/formation/{worker_type}"
-        headers = {
+        self.worker_type = settings.WORKER_DYNO_NAME
+        self.url = (
+            f"{self.API_ROOT}/{settings.HEROKU_APP_NAME}/formation/{self.worker_type}"
+        )
+        self.headers = {
             "Accept": "application/vnd.heroku+json; version=3",
             "Authorization": f"Bearer {settings.HEROKU_TOKEN}",
         }
```
davisagli: Since you're turning worker_type, url, and headers into instance attributes, it would make more sense to set them in __init__. That avoids the potential for something to try to use them when they haven't been set up yet.

config/settings/common.py
```diff
@@ -359,6 +364,9 @@ def url_prefix_list(val: str) -> List[str]:
 METACI_WORKER_RESERVE = env.int("METACI_WORKER_RESERVE", 1)
 WORKER_DYNO_NAME = env("WORKER_DYNO_NAME", default=None) or "worker"
 
+# Queues that the given app will has
+HEROKU_APP_QUEUES = env.list("HEROKU_APP_QUEUES", default=["high", "medium", "default"])
```
davisagli: This is no longer used

config/settings/local.py
```diff
@@ -123,6 +123,16 @@
 CONNECTED_APP_CLIENT_SECRET = env("CONNECTED_APP_CLIENT_SECRET")
 CONNECTED_APP_CALLBACK_URL = env("CONNECTED_APP_CALLBACK_URL")
 
+METACI_APPS = {"local-app"}
```
davisagli: Is there ever a reason for this to be different from the keys of the AUTOSCALERS dict? Can we just use those keys to avoid the possibility of them being accidentally out of sync?

config/settings/local.py
```diff
@@ -123,6 +123,16 @@
...
+
+AUTOSCALERS = {
+    "local-app": {
+        "app_name": "local-app",
+        "worker_type": "worker",
+        "queues": ["default", "medium", "high"],
```
davisagli: For running locally we should probably include the robot queue here (i.e. run it in the same workers locally)

config/settings/production.py
```diff
@@ -255,9 +256,14 @@
 
 HEROKU_TOKEN = env("HEROKU_TOKEN", default=None)
 HEROKU_APP_NAME = env("HEROKU_APP_NAME", default=None)
+
 if HEROKU_TOKEN and HEROKU_APP_NAME:
     METACI_WORKER_AUTOSCALER = "metaci.build.autoscaling.HerokuAutoscaler"
 
+METACI_APPS = env("METACI_APPS", default=None)
+# Autoscalers are defined per METACI_APP
+AUTOSCALERS = json.loads(env("AUTOSCALERS", default=None))
```
davisagli: For backwards compatibility, if the environment variable isn't set, let's construct a default AUTOSCALERS dict that uses the old HEROKU_APP and HEROKU_WORKER_NAME vars and old default queues.

config/settings/production.py
```diff
@@ -260,15 +261,19 @@
...
+if not AUTOSCALERS:
+    # backwards compatability
+    HEROKU_APP_NAME = env("HEROKU_APP_NAME", default=None)
+    AUTOSCALERS = {
+        HEROKU_APP_NAME: {
+            "app_name": HEROKU_APP_NAME,
+            "worker_type": "worker",
+            "queues": ["default", "medium", "high"],
+        }
+    }
```
davisagli: @Br4nd0R If HEROKU_APP_NAME is None, we should set AUTOSCALERS to an empty dict. (i.e. don't autoscale)

config/settings/local.py
```diff
@@ -123,6 +123,15 @@
 CONNECTED_APP_CLIENT_SECRET = env("CONNECTED_APP_CLIENT_SECRET")
 CONNECTED_APP_CALLBACK_URL = env("CONNECTED_APP_CALLBACK_URL")
 
+
+AUTOSCALERS = {
+    "local-app": {
+        "app_name": "local-app",
+        "worker_type": "worker",
```
davisagli: app_name and worker_type aren't actually needed for the base Autoscaler class, right?

config/settings/production.py
```diff
@@ -255,9 +257,24 @@
 
 HEROKU_TOKEN = env("HEROKU_TOKEN", default=None)
 HEROKU_APP_NAME = env("HEROKU_APP_NAME", default=None)
+
 if HEROKU_TOKEN and HEROKU_APP_NAME:
     METACI_WORKER_AUTOSCALER = "metaci.build.autoscaling.HerokuAutoscaler"
 
+# Autoscalers are defined per METACI_APP
+AUTOSCALERS = json.loads(env("AUTOSCALERS", default={}))
```
davisagli:
```suggestion
AUTOSCALERS = json.loads(env("AUTOSCALERS", default="{}"))
```

The default for `env` needs to be a string so that it can be loaded as JSON.

config/settings/production.py
```diff
@@ -255,9 +257,24 @@
...
+
+if not AUTOSCALERS:
+    # backwards compatability
+    HEROKU_APP_NAME = env("HEROKU_APP_NAME", default=None)
+    AUTOSCALERS = {
+        HEROKU_APP_NAME: {
+            "app_name": HEROKU_APP_NAME,
+            "worker_type": "worker",
```
davisagli:
```suggestion
            "worker_type": WORKER_DYNO_NAME,
```

for backwards-compatibility in case someone changed that setting

config/settings/local.py
Br4nd0R: Yes, this is correct.

README.rst
```diff
@@ -140,15 +140,19 @@ Click the bell icon at the top to view the My Notifications page (/notifications
 Automatic Scaling
 -----------------
 
-MetaCI can be configured to monitor its own build queue and scale its own Heroku dynos based on load. It will check the queue once a minute and add worker dynos when needed. Once all builds are complete, all worker dynos will be shut down. Heroku only bills for the dyno seconds used, so this scaling can save money while allowing for greater concurrency when desired.
+MetaCI can be configured to monitor its own build queue and scale its own Heroku dynos based on load in multiple Heroku Apps. It will check the queue once a minute and add worker dynos when needed. Once all builds are complete, all worker dynos will be shut down. Heroku only bills for the dyno seconds used, so this scaling can save money while allowing for greater concurrency when desired.
 
 To configure autoscaling:
 
 1. Set the METACI_MAX_WORKERS setting to the maximum number of dynos you'd like to scale up to.
 2. Set the METACI_WORKER_RESERVE setting to the number of dynos you'd like to reserve for high-priority builds. (Optional; defaults to 1.)
```
davisagli: @Br4nd0R reading this makes me think we should move the max workers and worker reserve settings to be part of the autoscaler config dict too. We don't necessarily want the same max for the robot app, and it doesn't need a reserve.

README.rst
```diff
@@ -140,15 +140,19 @@ Click the bell icon at the top to view the My Notifications page (/notifications
...
 3. Set up a Heroku user with access to this app, and create an authorization token using ``heroku authorizations:create``. Set the HEROKU_TOKEN setting to this authorization token.
 4. Set the HEROKU_APP_NAME setting to the name of the Heroku app.
```
davisagli: We can remove this line; AUTOSCALERS takes its place.

PR #609, opened 2019-11-15
Fix equality test in plan handler

jstvz:
----

#


jstvz:
Just saw it fixed in 09b5295

PR #634, opened 2019-11-20
update AUTOSCALERS

Br4nd0R:


config/settings/local.py
```diff
@@ -124,7 +124,7 @@
 CONNECTED_APP_CALLBACK_URL = env("CONNECTED_APP_CALLBACK_URL")
 
 
-AUTOSCALERS = {"local-app": {"queues": ["default", "medium", "high", "robot"],}}
+AUTOSCALERS = {"local-app": {"max_workers": 1, "worker_reserve": 0, "queues": ["default", "medium", "high", "robot"],}}
```
davisagli: This is a reasonable default, but it'd be nice to still support overriding via an env var locally

PR #635, opened 2019-11-21
Add task field to TestResult model; update importer to use it.

boakley:
This adds `robot_task` as a field on the TestResult mode, so that we can access the options on the
task when regenerating the log.


metaci/testresults/models.py
```diff
@@ -1,24 +1,21 @@
...
 from django.utils import timezone
-from django.apps import apps
-
 from postgres_stats.aggregates import Percentile
 
-from metaci.testresults.choices import OUTCOME_CHOICES
-from metaci.testresults.choices import TEST_TYPE_CHOICES
-
 from metaci.build import models as build_models
+from metaci.testresults.choices import OUTCOME_CHOICES, TEST_TYPE_CHOICES
```
davisagli: Did you do this, or was it some automated import sorting in your editor? The rationale behind keeping them on separate lines is to avoid noise like this in diffs.

metaci/testresults/models.py
```diff
@@ -116,6 +113,18 @@ class TestResult(models.Model):
...
+    # test was run (e.g. noncritical, tagstatlink, etc)
+    robot_task = models.ForeignKey(
+        "build.FlowTask",
+        related_name="test_results",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        default=None,
+        help_text="Task which generated the original output.xml file",
+    )
```
davisagli: The TestResult model is for all test results, not just robot tests. Can we name this field `task` instead of `robot_task` to leave open the possibility of linking non-robot test results to their task as well? (It's fine if we leave actually doing that for later; I just don't see a reason to name this in a way that limits it to being for robot tests.)

metaci/testresults/models.py
boakley: I don't recall intentionally doing it.

metaci/testresults/models.py
boakley: I'm glad you brought that up. It's tempting to name it `task`, but the code only sets it for robot tasks. I don't think there's any way to know how to set it for other records. The way I set it now is I look at the timestamp of the output.xml file and then find a task that started and ended around that time. Without an output.xml file I don't know how to set it. If there's a better way for a specific task to associate itself with some test results, I'm all ears. 

I have no problem renaming it `task` if you think that makes more sense. I can just add a comment that says says it's currently only valid for robot tasks.

boakley: I was tempted to create a separate model specifically for robot tasks and add a foreign key to it in the TestResult mode -- we could use it to add a lot of robot-specific statistics. That seemed like too big of a change at this time, especially for my first foray into the metaci code base, and especially because I don't have my head wrapped around the whole model.

metaci/conftest.py
```diff
@@ -207,3 +212,12 @@ class Meta:
...
     branch = factory.SubFactory(BranchFactory)
     plan = factory.SubFactory(PlanFactory)
+
+
+class FlowTaskFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = FlowTask
+
+    build_flow = factory.SubFactory(BuildFlowFactory)
+    stepnum = factory.Sequence(lambda n: "{}.".format(n))
```
Br4nd0R:
```suggestion
    stepnum = factory.Sequence(lambda n: f"{n}.")
```

metaci/conftest.py
```diff
@@ -207,3 +212,12 @@ class Meta:
...
+    path = factory.Sequence(lambda n: "flow_1.task_{}".format(n))
```
Br4nd0R:
```suggestion
    path = factory.Sequence(lambda n: f"flow_1.task_{n}")
```

metaci/testresults/models.py
davisagli: I agree adding a separate model for robot test results feels like a big change to tackle right now.

I like the idea of calling it `task` and adding a comment that it's currently only set for robot tests.

boakley: I renamed the field to `task` and added a brief comment

metaci/testresults/robot_importer.py
```diff
@@ -13,6 +15,20 @@ def import_robot_test_results(build_flow, path):
...
+    path = pathlib.Path(path)
+
+    # find the robot task that generated the log file so
+    # that we can access the task options when rendering
+    # the log file
+    mtime = timezone.make_aware(
+        datetime.fromtimestamp(path.stat().st_mtime), timezone.get_current_timezone()
+    )
+    robot_task = None
+    for task in build_flow.tasks.order_by("-time_end"):
```
davisagli: you could add `.filter(class_path="Robot")` to only get the robot tasks. (It should be `cumulusci.tasks.robotframework.Robot` but it's not getting set correctly here https://github.com/SFDO-Tooling/MetaCI/blob/master/metaci/build/flows.py#L18)

metaci/testresults/robot_importer.py
```diff
@@ -13,6 +15,20 @@ def import_robot_test_results(build_flow, path):
...
+        if task.time_start <= mtime <= task.time_end:
```
davisagli: At first I thought doing this based on the timestamp felt brittle, but I guess it has the advantage of picking the right task if multiple tasks wrote to the same location.

metaci/conftest.py
```diff
@@ -207,3 +212,12 @@ class Meta:
...
     branch = factory.SubFactory(BranchFactory)
     plan = factory.SubFactory(PlanFactory)
+
+
+class FlowTaskFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = FlowTask
+
+    build_flow = factory.SubFactory(BuildFlowFactory)
+    stepnum = factory.Sequence(lambda n: f"{n}.")
```
davisagli: Why'd you include the period here? Normal stepnums would like one of these:

    1  # a task that is the 1st step at the top level
    1.1  # a task that is the 2nd step at the top level (sorted numerically within a level)
    2/1  # task which is the first step within a flow that is the second step at the top level

so they can include periods, but it's odd to have one that's not being used as a separator

metaci/testresults/views.py
```diff
@@ -171,16 +171,63 @@ def test_result_robot(request, result_id):
...
+    """
+    # Yeah, I know patching HTML is fraught with peril. The robot
+    # code to generate the logs is pretty stable, so I think
+    # this is a reasonably safe way to do it. It results in a
+    # much better experience for our users.
+    html = html.replace(
+        r'<span>[<a href="{{html $value.url}}" title="{{html $value.url}}">',
+        r'<span>[<a href="{{html $value.url}}" title="{{html $value.url}}" target="_top">',
+    )
+    return html
```
davisagli: What about just adding `<base target="_top">` to the `<head>`?

metaci/conftest.py
boakley: hmmm. I don't recall why I did that. Might have been my ignorance about step numbers. I'll take a look.

metaci/testresults/views.py
boakley: I started to do that, but there are other links on the page that definitely shouldn't open in a new window. For example, collapsible lines have little icons that will expand or collapse all the children. If we set the base target, those completely break. 

I think this is the safest change at the moment since it only affects the new functionality and doesn't change any existing behavior.

davisagli: @boakley oh good point

metaci/conftest.py
boakley: I couldn't find any reason for that change so I reverted it. You win the eagle-eye award for catching this!

metaci/testresults/robot_importer.py
boakley: I suppose you're right, but the filter doesn't really buy us much here since there will likely only ever be a handful of tasks and we're often going to find the one we're looking for very early in the loop.

PR #690, opened 2019-12-03
📦 flask app for redirecting requests to new metaci app

Br4nd0R:
* I think we would need to add the FLASK_APP config var to `redirects.py`


davisagli:
@Br4nd0R or we can serve the app using gunicorn like this: https://github.com/SFDO-Tooling/mrbelvedere/blob/master/Procfile

davisagli:
Closing this -- we are using it in the old mrbelvedereci app, but it should not be merged to master.

PR #695, opened 2019-12-04
Coverage for metaci.testresults.view.py

Br4nd0R:
# Changes
* We now have pytest fixtures (`data`, `user`, and `superuser`) for sample data that are available for use across ALL test classes in MetaCI (see `metaci.fixtures.data_fixtures.py`).
* The `data` fixture contains a dict of related objects that might need to be referenced in multiple different contexts. If there are generic relationships that don't exist they should be added to the `data` fixture in the future. For special—one off—changes that need to exist only for a single test, they should be created in the test itself.


metaci/testresults/tests/conftest.py
```diff
@@ -0,0 +1 @@
+from metaci.fixtures.data_fixtures import *
```
davisagli: Why not just make the data_fixtures globally available from the top-level conftest.py? (That'll probably require moving the factories to a separate module to avoid import cycles.)

metaci/testresults/tests/test_testresults_views.py
```diff
@@ -0,0 +1,117 @@
+import pytest
+from django.test import Client
+from django.urls import reverse
+
+from metaci.testresults import views
+
+
+class TestTestCoverageViews:
```
davisagli:
```suggestion
class TestTestResultsViews:
```

metaci/testresults/tests/test_testresults_views.py
```diff
@@ -0,0 +1,117 @@
+import pytest
```
davisagli: I would call this file test_views.py. I try to follow the pattern that if the actual code is in somepackage/somemodule.py, then the corresponding tests are in somepackage/tests/test_somemodule.py

(Oddbird goes even further and puts them in somepackage/tests/somemodule.py without the test_ prefix, but then it's trickier to make pytest find the right files so I have mixed feelings about it.)

metaci/testresults/tests/test_testresults_views.py
```diff
@@ -0,0 +1,117 @@
...
+        self.client.force_login(superuser)
+        url = reverse("test_method_trend", kwargs={"method_id": data["testmethod"].id})
+        response = self.client.get(url)
+
+        assert response.status_code == 200
+
+    @pytest.mark.skip
+    @pytest.mark.django_db
+    def test_build_flow_compare__superuser(self, data, superuser):
+        # TODO: NoReverseMatch troubleshooting
```
davisagli: The NoReverseMatch is happening when build_flow_compare.html tries to do a reverse lookup for "commit_detail" using execution1.build.commit which is empty

metaci/testresults/tests/test_testresults_views.py
```diff
@@ -0,0 +1,117 @@
...
+            {"buildflow1": data["buildflow"].id, "buildflow2": data["buildflow"].id,},
+        )
+        assert response.status_code == 200
+
+    @pytest.mark.skip
+    @pytest.mark.django_db
+    def test_test_dashboard__superuser(self, data, superuser):
+        """Currently failing because test_dashboard has a fileter
+        that references a testresult field on TestMethod objs
+        have no clue what the code is trying to do."""
```
davisagli: I think this view is one that Christian started working on but didn't finish. It's not working in production either. Let's remove it.

metaci/testresults/tests/conftest.py
Br4nd0R: I had played around with that first, but could never get it to work. Maybe because of the factories being in the top-level conftest.py?

metaci/testresults/tests/test_testresults_views.py
Br4nd0R: Sounds good to me.

PR #699, opened 2019-12-06
Feature/robot docker

jofsky:
Creating Chrome and Chromedriver package installation steps in Dockerfile


davisagli:
I agree with most of @boakley's feedback, but not this one:

> There are several places where you COPY a helper script and then RUN the helper script. Instead of COPY,RUN,COPY,RUN, ... would it make more sense to have a single COPY at the start which copies all of the helper scripts at once?

A docker image is composed of multiple layers. Each RUN command creates a layer which can be reused as long as the files that were copied before it have not changed. Moving all the copies to the beginning would invalidate all the layers whenever any file changes, so rebuilding the image would take longer.

Dockerfile
```diff
@@ -12,7 +14,13 @@ ENV YARN_VERSION 1.19.1
 # installing yarn
 COPY ./utility/install_yarn.sh /app/utility/install_yarn.sh
 RUN /bin/sh /app/utility/install_yarn.sh
-# # installing sfdx
+# Installing Google Chrome
+COPY ./utility/install_chrome.sh /app/utility/install_chrome.sh
+RUN /bin/sh /app/utility/install_chrome.sh
+# Installing Chromedriver
+COPY ./utility/install_chrome.sh /app/utility/install_chrome.sh
```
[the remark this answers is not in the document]
jofsky: Good catch.

.gitignore
```diff
@@ -101,4 +101,6 @@ postgres/
 # safeguarding secrets
 sfdx_hub.key
 
-docs/_build 
\ No newline at end of file
+docs/_build
+output.pdf
```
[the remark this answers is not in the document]
jofsky: please take a look

utility/wrap_chrome_binary
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
jofsky: please take a look

utility/install_chromedriver.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash 
+
```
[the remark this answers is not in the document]
jofsky: please take a look

utility/install_chromedriver.sh
```diff
@@ -0,0 +1,15 @@
...
+then 
+    CHROME_MAJOR_VERSION=$(google-chrome --version | sed -E "s/.* ([0-9]+)(\.[0-9]+){3}.*/\1/")
+    CHROME_DRIVER_VERSION=$(wget --no-verbose -O - "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_${CHROME_MAJOR_VERSION}");
+fi
+echo "Using chromedriver version: ${CHROME_DRIVER_VERSION}"
+wget --no-verbose -O /tmp/chromedriver_linux64.zip https://chromedriver.storage.googleapis.com/$CHROME_DRIVER_VERSION/chromedriver_linux64.zip
+rm -rf /opt/selenium/chromedriver 
+unzip /tmp/chromedriver_linux64.zip -d /opt/selenium 
+rm /tmp/chromedriver_linux64.zip 
+mv /opt/selenium/chromedriver /opt/selenium/chromedriver-$CHROME_DRIVER_VERSION
```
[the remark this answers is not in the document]
jofsky: Im not sure this is necessary I was able to achieve success without linking the PATH on my container. Lets discuss tomorrow.

utility/install_chrome.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash 
+
```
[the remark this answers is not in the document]
jofsky: please take a look

Dockerfile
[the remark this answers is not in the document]
jofsky: So this leads me to question if we need this step?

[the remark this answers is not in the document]
boakley: I don't know. We definitely need chromedriver. I don't know if `install_chrome.sh` also installs chromedriver or not. I suspect it does not.

utility/install_chrome.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash 
 
+# This file installs the given CHROME_VERSION to the docker container downloading 
+# a stable version to the dockerfile image
```
boakley: I'm finding this hard to grok. Wouldn't it be better to say that it installs chrome rather than CHROME_VERSION? I also don't understand what "downloading a stable version to the dockerfile image" means. It sounds like it will install the version specified by CHROME_VERSION _and_ a stable version, which isn't what it's doing. Maybe you can reword this to say it's installing chrome, and choosing  either the stable version, or the version defined by CHROME_VERSION.

utility/install_chromedriver.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash 
 
+# This file installs the given CHROME_DRIVER_VERSION to the docker
+# container downloading a stable version to the dockerfile image
+# If no CHROME_DRIVER_VERSION is declared it will take the local systems
+# CHROME VERSIONS.
+
```
boakley: same comment as I made in install_chrome.sh. I think this needs a bit of rewording. The comment sounds like it's installing whatever is in CHROME_DRIVER_VERSION and a stable version. Plus, it's not installing CHROME_DRIVER_VERSION, it's installing chromedriver, with the version defined by CHROME_DRIVER_VERSION.

utility/start_server.sh
```diff
@@ -6,7 +6,7 @@ python /app/manage.py migrate
 if [ "${DJANGO_SETTINGS_MODULE}" = "config.settings.local" ] ; then
     echo "CREATING ADMIN USER FOR TESTING PURPOSES..."
     # Using key error as an indicator for whether or not to run database population and job scheduling scripts
-    echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', 'admin@salesforce.com', 'password')" | python manage.py shell
+    echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', 'admin@test.com', 'password')" | python manage.py shell
```
boakley: I recommend using `example.com` here. `test.com` is an actual real-live domain owned by somebody, and we wouldn't want a bug to cause them to get unwanted email. 

`example.com`, on the other hand, is specifically designed to be used in documentation and code examples.

boakley:
> I agree with most of @boakley's feedback, but not this one:
> 
> > There are several places where you COPY a helper script and then RUN the helper script. Instead of COPY,RUN,COPY,RUN, ... would it make more sense to have a single COPY at the start which copies all of the helper scripts at once?
> 
> A docker image is composed of multiple layers. Each RUN command creates a layer which can be reused as long as the files that were copied before it have not changed. Moving all the copies to the beginning would invalidate all the layers whenever any file changes, so rebuilding the image would take longer.

You make a good point. I must have had my "pretty code" hat on, rather than my "efficient code" hat on.  I blame my suggestion on the hats.

Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN /bin/sh /app/utility/install_yarn.sh
 COPY ./utility/install_chrome.sh /app/utility/install_chrome.sh
 RUN /bin/sh /app/utility/install_chrome.sh
 # Installing Chromedriver
-COPY ./utility/install_chrome.sh /app/utility/install_chrome.sh
+COPY ./utility/install_chromedriver.sh /app/utility/install_chromedriver.sh
```
boakley: You need to modify the following line too. You're copying `install_chromedriver.sh` but you're executing `install_chrome.sh`.

utility/start_server.sh
jofsky: GREAT catch jason said the same!

Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG BUILD_ENV
 ARG CHROME_VERSION
 ARG CHROMEDRIVER_VERSION
 
-RUN mkdir /app
+RUN mkdir /app; mkdir /app/.apt/; mkdir /app/.apt/usr/;mkdir /app/.apt/usr/bin/;
```
boakley: You can use the `-p` option of mkdir to make the entire path in a single command:

    mkdir -p /app/.apt/usr/bin

boakley: Also, now that a local .apt folder is not needed, it should be removed from the repository.

boakley:
A couple of lingering thoughts:

- First, this is cool. I know it was a headache to create, but it's a giant step in the right direction. 👍 
- it would be nice if the ports were configurable with arguments or environment variables. If I already have some other tool using port 8000 (or am running metaci locally), it would be nice to tell docker to use a different set of ports.
- this might be a good time to write a simple smoke test in robot.  It would be reassuring to be able to do `docker-compose up` followed by `robot src/test/robot/smoke-test.robot` to have it open a browser, go to the landing page, and maybe go to the plans page to verify all of the pre-configured plans are there.

CONTRIBUTING.rst
```diff
@@ -96,6 +89,21 @@ translations to ``locales/<language>/translation.json``.
...
+
+If you need to use libraries that do not have flow definitions, you could edit
+a file with a name like ``flow-typed/npm/@package/module_vx.x.x.js`` to stub out addition component
+type definitions. OR you can run ``flow-typed update --ignoreDeps dev`` to allow
+it to automatically generate stubs for modules with missing type definitions.
+
+At some point we will probably move to TypeScript.
+
+.. _flow: https://flow.org/
+
```
davisagli: @jofsky how did this section get added here? we switched MetaCI to TypeScript a while ago so this info is not accurate.

config/settings/production.py
```diff
@@ -1,301 +1,404 @@
...
-SFDX_CLIENT_ID = env("SFDX_CLIENT_ID")
-SFDX_HUB_KEY = env("SFDX_HUB_KEY")
-SFDX_HUB_USERNAME = env("SFDX_HUB_USERNAME")
+# django-guardian settings
+GUARDIAN_MONKEY_PATCH = False
 
-# django-defender configuration
-DEFENDER_REDIS_NAME = "default"
+JS_REVERSE_JS_VAR_NAME = "api_urls"
+JS_REVERSE_EXCLUDE_NAMESPACES = ["admin", "admin_rest"]
```
davisagli: @jofsky why are there so many changes in production.py? I would expect to see minimal changes here since we are not switching production to use docker at this time.

docker/utility/start_server.sh
```diff
@@ -6,7 +6,7 @@ python /app/manage.py migrate
 if [ "${DJANGO_SETTINGS_MODULE}" = "config.settings.local" ] ; then
     echo "CREATING ADMIN USER FOR TESTING PURPOSES..."
     # Using key error as an indicator for whether or not to run database population and job scheduling scripts
-    echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', 'admin@salesforce.com', 'password')" | python manage.py shell
+    echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', '$SFDX_HUB_USERNAME', 'password')" | python manage.py shell
```
davisagli: I would simply set this to admin@example.com. There's no guarantee that a Salesforce username is a deliverable email address.

docker/utility/wrap_chrome_binary.sh
```diff
@@ -0,0 +1,12 @@
...
+WRAPPER_PATH=$(readlink -f /usr/bin/google-chrome)
+BASE_PATH="$WRAPPER_PATH-base"
+mv "$WRAPPER_PATH" "$BASE_PATH"
+
+cat > "$WRAPPER_PATH" <<_EOF
+#!/bin/bash
+# Note: exec -a below is a bashism.
+exec -a "\$0" "$BASE_PATH" --no-sandbox "\$@"
+_EOF
+chmod +x "$WRAPPER_PATH"
```
davisagli: What is this wrapper for?

package.json
```diff
@@ -12,6 +12,8 @@
     "@salesforce/design-system-react": "^0.10.16",
     "@salesforce/lwc-jest": "^0.6.1",
     "@sentry/browser": "^5.10.2",
+    "chrome": "^0.1.0",
+    "chromedriver": "^78.0.1",
```
davisagli: These should not be here. They are installed using the commands in the Dockerfile, not as node packages.

docker/utility/wrap_chrome_binary.sh
jofsky: allows for headless running of google-chrome for selenium and robot testing, getting an error without it being run in Dockerfile

davisagli: Hmm, I think we should look for a way to let chrome run with the sandbox feature turned on. For one thing, that will make it more similar to the other environments where we run robot tests. For another, turning off the sandbox sounds non-ideal since it's turning off a security feature, and a bit of googling suggests it is discouraged.

What was the error you were getting?

davisagli: https://ndportmann.com/chrome-in-docker/ discusses an approach that sounds better

boakley: IIRC, this solves the problem that we needed (or thought we needed) the `--no-sandbox` option and we didn't want to modify our open_test_browser keyword in order to add it.

What I don't know is if we truly require that option. Some internet advice said we did, but I think that was advice from several years ago. I _think_ (but don't know for certain) that it's only required if you're running as root, which I think the docker container is doing. 

We probably should be creating a user inside the docker container instead of running as root. That might make the need for the option to go away, and with that we would no longer need the wrapper.

jofsky: Screen Shot 2019-12-30 at 5 05 14 PM

davisagli: Having dug into this for a while, I think adding `--no-sandbox` like this is the most pragmatic option. The container environment already provides one level of sandboxing. Sandboxing chrome from the rest of the container would be more critical if we were running tests from an untrusted source. We'll need to revisit this if we want to provide more isolation between MetaCI itself and the builds that it is running.

docker/utility/start_server.sh
```diff
@@ -25,4 +25,9 @@ echo "AUTHORIZING SFDX via JWT key auth..."
 sfdx force:auth:jwt:grant -u $SFDX_HUB_USERNAME -f /app/sfdx_hub.key -i $SFDX_CLIENT_ID --setdefaultdevhubusername
 # starting server and webpack
 yarn serve
-echo "ALL DONE!"
+
+echo "*****************************************************
+MetaCI is up and running in docker. 
+
+Go to http://localhost:8000/admin/login to log in
+******************************************************"
```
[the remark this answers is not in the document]
jofsky: These errors are cause the keys are already created in the db? Is this a dealbreaker/how is the best way to handle this while still having meaningful messaging?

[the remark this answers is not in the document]
davisagli: There are lots of reasons there could be an error.

I think I would just remove this message. when it runs successfully, `yarn serve` runs continually and does not exit, so it won't actually print the message in the situation where it would be correct.

config/settings/production.py
jofsky: So i had to use --no-verify on the commit to get it to pass with the current production.py file

davisagli: That's normal -- since it uses "import *", it's not compliant with our flake8 checks. So we usually just manually review to make sure there aren't _other_ flake8 errors, and then check it in with --no-verify.

There's probably a better way, but replacing all of production.py is DEFINITELY not a good solution. Extending from a common set of base settings is a pattern we use in all our projects that has worked well for us. I'm willing to discuss ways we might change that, but let's consider it separately from the docker changes for robot.

PR #700, opened 2019-12-06
Feature/brandon/moaaar coverage

Br4nd0R:


.gitignore
```diff
@@ -64,6 +65,7 @@ node_modules/
 .env
 .envrc
 local_setup.sh
+list/
```
davisagli: What is this?

PR #726, opened 2019-12-09
Test coverage for metaci.plan.views

Br4nd0R:


metaci/plan/tests/test_views.py
```diff
@@ -0,0 +1,102 @@
...
+import pytest
+from django.test import Client
+from django.urls import reverse
+
+from metaci.fixtures.factories import BranchFactory
+
+
+@pytest.mark.django_db
+class TestPlanViews:
+    client = Client()
```
davisagli: This should be created in a fixture once per test rather than shared between test cases. There's too much risk of something like a cookie accidentally leaking between tests.

Br4nd0R: I went ahead and made a new `fixtures/util.py` module for holding the `client` fixture. I'll need to circle back at some point and make sure it's used across all test classes where I made the test client an attribute of the test class.

PR #728, opened 2019-12-13
add a happy path test of Build.run

davisagli:


metaci/build/tests/test_models.py
```diff
@@ -48,10 +48,40 @@ def test_planrepo_find_on_build_init(self):
...
+        get_org.return_value = org_config
+
+        build = BuildFactory()
+        build.plan.flows = "test"
+
+        try:
+            build.run()
+        finally:
+            detach_logger(build)
+            os.chdir(cwd)
```
Br4nd0R: If this is the only place the `cwd` variable is being referenced I think it would be a tiny bit better if the variable declaration was closer to this location.

davisagli: I actually want to switch to an autouse fixture that resets the cwd after each test -- like this but without the class scope: https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/conftest.py#L32

davisagli: @Br4nd0R done and ready to re-review

PR #753, opened 2019-12-16
🐈 Dependency Updates

Br4nd0R:


davisagli:
@Br4nd0R looks like we need to hold back `coverage` here too -- note the error reporting to coveralls in the Heroku CI log and the fact that we didn't get a commit status from coveralls on this PR

PR #779, opened 2019-12-23
Install Snowfakery optionally

prescod:
Install Snowfakery only if the SFDO_METACI_DEPLOY environment variable is set.

Supply username and password with environment variables.

Suppress output for security reasons.

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -1,0 +1,6 @@
+echo SFDO_METACI_DEPLOY=${SFDO_METACI_DEPLOY}
+if [[ -z "${SFDO_METACI_DEPLOY}" ]]; then
+    echo Installing from sfdo-metaci.txt
+    pip3 install -r requirements/sfdo-metaci.txt > /dev/null 2>&1 || echo Pip install failed. Error suppressed for security reasons
+fi
+echo "Ran post_compile hook"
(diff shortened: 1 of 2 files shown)


bin/post_compile
```diff
@@ -0,0 +1,6 @@
+echo SFDO_METACI_DEPLOY=${SFDO_METACI_DEPLOY}
+if [[ -z "${SFDO_METACI_DEPLOY}" ]]; then
+    echo Installing from sfdo-metaci.txt
+    pip3 install -r requirements/sfdo-metaci.txt > /dev/null 2>&1 || echo Pip install failed. Error suppressed for security reasons
```
davisagli: Does it include the github username & access token in the log otherwise? Since this is suppressed, did you validate that snowfakery was successfully installed in some other way?

prescod: > Does it include the github username & access token in the log otherwise?

Yes.

> did you validate that snowfakery was successfully installed in some other way?

When we get to staging I will log into a dyno and verify.

davisagli: Okay. I wonder if we need to add `#egg=snowfakery` to the end of the git URL so that pip knows what package to install, but I guess we'll find out.

davisagli:
@prescod don't forget to set SFDO_METACI_DEPLOY on metaci-stg and metaci-robot-stg before you merge. (metaci-robot-stg is a separate app that will soon be used for running robot builds, but isn't active yet) And on sfdo-metaci and sfdo-metaci-robot before we promote to production (but keep in mind that that will restart them, so the usual rules about waiting for them to be idle still apply)

davisagli:
@prescod er, a correction: metaci-robot-stg is already used for running robot builds from the staging environment. But we haven't activated the corresponding production app, sfdo-metaci-robot, yet.

PR #790, opened 2019-12-27
Move pip upgrade from post_compile to pre_compile

prescod:
Move pip upgrade from post_compile to pre_compile so it is available earlier in the process.

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -1,6 +1,4 @@
 echo SFDO_METACI_DEPLOY=${SFDO_METACI_DEPLOY}
-pip install --upgrade pip    # Upgrade pip to get embeddable env vars feature
-pip --version
 if [ -n "${SFDO_METACI_DEPLOY}" ]; then
     echo Installing from sfdo-metaci.txt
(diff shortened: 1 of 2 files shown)


davisagli:
The buildpack is installing its own version of pip _after_ the pre-compile hook. So then the post-compile hook fails like this:

```
-----> Running post-compile hook
SFDO_METACI_DEPLOY=SFDO_METACI_DEPLOY
Installing from sfdo-metaci.txt
Command "git clone -q https://${GITHUB_USERNAME}:${GITHUB_PASSWORD}@github.com/SalesforceFoundation/Snowfakery.git /tmp/pip-m4gxk70p-build" failed with error code 128 in None
You are using pip version 9.0.2, however version 19.3.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
/app/.heroku/python/bin/python: Error while finding module specification for 'snowfakery.cli' (ModuleNotFoundError: No module named 'snowfakery')
Warning: Snowfakery is not installed. If you don't use it, that is not a problem.
Ran hook from bin/post_compile
```

davisagli:
Closing this since it looks like it won't work after all.

PR #816, opened 2020-01-11
Feature/robot docker

jofsky:


davisagli:
@jofsky we already merged the feature/robot-docker branch. What's the purpose of these changes? I'm also confused why the PR shows so many commits. If there are one or two tweaks needed, it would probably make more sense to start a new branch from current master.

PR #841, opened 2020-01-14
Updating refactored vscode docker plugin file extensions

jofsky:


davisagli:
@jofsky What are you trying to accomplish here?

jofsky:
I wanted to keep the docker-compose.dev.yml naming convention like MetaShare's

davisagli:
@jofsky ah ok, just to give a clue that it's meant for development? I'm fine with that, but it looks like this PR removes docker-compose.yml without adding docker-compose.dev.yml

jofsky:
Ok @davisagli  should be fixed!

.devcontainer/docker-compose.dev.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+
+services:
+  web:
+    volumes:
+      - .:/app:delegated
+      - vscode-server:/root/.vscode-server:cached
+      - /var/run/docker.sock:/var/run/docker.sock:cached
```
davisagli: @jofsky This line is new. Does adding it solve a problem?

jofsky: No I don't believe so let me try removing see what happens.

jofsky:
@davisagli  I had included since it was added in MetaShare but it doesn't seem to be required do you know why it was added in that docker-compose.dev.yml?

davisagli:
@jofsky got it. I wasn't sure about that one but I just searched and think it might be a solution for this: https://code.visualstudio.com/docs/remote/faq#_why-cant-i-see-my-local-containers-in-the-docker-extension-when-using-the-remote-extensions

However since we haven't seen that problem in MetaCI, I'd prefer to leave it out.

davisagli:
@jofsky the other thing that I'm trying to figure out here is all the whitespace differences that github is showing in devcontainer.json. I looked at the file in a hex editor and it looks like all the spaces got turned into non-breaking spaces.

jofsky:
@davisagli I think it may be due to my vscode settings

davisagli:
@jofsky thanks for cleaning that up

PR #869, opened 2020-01-21
🛢 Make plan lock timeout configurable.

prescod:
The main thing this does is allow a plan to specify a build timeout instead having it be hard-coded to 8 hours.

A test was created to ensure it works.

Also: it seems like a previous migration was not created so makemigrations did that too.


metaci/plan/migrations/0033_auto_20200121_0110.py
```diff
@@ -0,0 +1,32 @@
...
+        ("plan", "0032_set_plan_queues"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="plan",
+            name="build_timeout",
+            field=models.IntegerField(default=28800),
+        ),
+        migrations.AlterField(
```
Br4nd0R: @prescod - Do you know why this migration says it wants to alter the `Plan.queue` field?

Br4nd0R: I would have guessed that it should only be adding the new build_timeout field.

prescod: Sorry I wasn't clear in the PR comment. Makemigrations created that because at some point in the past someone didn't run makemigrations after they changed a model. It can be a bit tricky to know what model changes will trigger a migration so it is easy to forget. 

We have discussed adding a validation in our build system for that, so you could add that to a todo list if you like.

There was actually another migration generated. I ignored it because it looked empty and wasn't in this file at all. I don't remember which model it was on but it looked empty.

prescod: Also: the code coverage went up when I did my commits but seemed to go down when I merged master. So I think that its master that went down, not my PR.

Br4nd0R: Ya I'm confused by the (-0.1%) in Coveralls (which I saw in my last MetaCI PR on Friday as well).

Br4nd0R: Would it be ok to hold off on merging this until I deploy to the metaci robot app tonight? (I want to ensure that both apps are on the same version and don't want to have to deploy to both tonight).

davisagli:
Good idea. I wonder if we should add validation to prevent setting this above 24 hours, since Heroku won't run dynos longer.

PR #909, opened 2020-02-03
Week 6 Dependency Updates

Br4nd0R:


davisagli:
@Br4nd0R It looks like there is now a 3.4.1 release of the Python redis client that should avoid the error. https://github.com/andymccurdy/redis-py/blob/master/CHANGES

PR #931, opened 2020-02-15
Capture CCI version in build log.

prescod:
Self-explanatory. Just output the version number to let QEs see what version is in play.

Proof that it works:

https://sfdo-metaci.herokuapp.com/builds/182785

### DIFF
--- a/metaci/build/models.py
+++ b/metaci/build/models.py
@@ -10,6 +10,7 @@ import zipfile
 from glob import iglob
 from io import BytesIO
 
+from cumulusci import __version__ as cumulsuci_version
 from cumulusci.core.config import FAILED_TO_CREATE_SCRATCH_ORG
 from cumulusci.core.exceptions import (
     ApexTestException,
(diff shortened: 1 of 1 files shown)


prescod:
The coverage complaint doesn't make any sense. Hopefully it won't delay this PR.

metaci/build/models.py
```diff
@@ -10,6 +10,7 @@
 from glob import iglob
 from io import BytesIO
 
+from cumulusci import __version__ as cumulsuci_version
```
davisagli:
```suggestion
from cumulusci import __version__ as cumulusci_version
```

metaci/build/models.py
```diff
@@ -288,7 +289,9 @@ def worker_id(self):
     def run(self):
         self.logger = init_logger(self)
         worker_str = f"in {self.worker_id}" if self.worker_id else ""
-        self.logger.info(f"-- Building commit {self.commit} {worker_str}")
+        self.logger.info(
+            f"-- Building commit {self.commit} {worker_str} with CumulusCI version {cumulsuci_version}"
```
davisagli:
```suggestion
            f"-- Building commit {self.commit} {worker_str} with CumulusCI version {cumulusci_version}"
```

PR #1045, opened 2020-03-31
Update Dockerfile

jstvz:
This PR fixes (🛠) some issues I encountered last week, and includes some improvements (✨) I discovered while learning more about how Docker works.

- 🛠 Update Dockerfile Yarn version

- ✨ Use multiline ENV: While ENV instructions don't create layers, they do create a a temporary layer that is squashed, and this takes time.

- 🛠 Use build argument defaults: When an argument and environment variable have the same name, the environment variable takes precedence. See [ENV]

- ✨ Pip install RUN doesn't use a conditional, and instead uses the `BUILD_ENV`
  argument to select which requirements.txt to use.

- ✨ Combine packaging commands into single RUN instruction. See [apt-get]
  - `--no-install-recommends`: Don't install libu2f-udev

- 🛠 Add postgres password
  Otherwise we'll get the following error:

```
postgres_1  | Error: Database is uninitialized and superuser password is not specified.
postgres_1  |        You must specify POSTGRES_PASSWORD to a non-empty value for the
postgres_1  |        superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".
postgres_1  |
postgres_1  |        You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
postgres_1  |        connections without a password. This is *not* recommended.
postgres_1  |
postgres_1  |        See PostgreSQL documentation about "trust":
postgres_1  |        https://www.postgresql.org/docs/current/auth-trust.html
mci_postgres_1 exited with code 1
```

[apt-get]: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

[ENV]: https://docs.docker.com/engine/reference/builder/#env#using-arg-variables

----

#


Dockerfile
```diff
@@ -47,7 +51,7 @@ RUN /bin/sh /app/docker/utility/install_sfdx.sh
 # installing python related dependencies with pip
 COPY ./requirements /app/requirements
 RUN pip install --no-cache --upgrade pip
-RUN if [ "${BUILD_ENV}" = "production" ] ; then pip install --no-cache -r /app/requirements/production.txt ; else pip install --no-cache -r /app/requirements/local.txt ; fi
+RUN pip install --no-cache -r /app/requirements/$BUILD_ENV.txt
```
davisagli: Nice simplification.

I just noticed, I think the flag for pip is supposed to be `--no-cache-dir` rather than `--no-cache`

Dockerfile
```diff
@@ -1,23 +1,27 @@
 FROM python:3.8
 
-ARG BUILD_ENV
-ARG CHROMEDRIVER_VERSION
+ARG BUILD_ENV=local
+ARG CHROMEDRIVER_VERSION=80.0.3987.106
 
 RUN mkdir -p /app/.apt/usr/bin
 
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
davisagli: What's this for?

Dockerfile
```diff
@@ -56,17 +60,14 @@ WORKDIR /app
...
-# in docker-compoose
-ENV REDIS_URL "redis://redis:6379"
-ENV DJANGO_SETTINGS_MODULE config.settings.local
-ENV DATABASE_URL postgres://metaci@postgres:5432/metaci
-ENV DJANGO_HASHID_SALT 'sample hashid salt'
-ENV DJANGO_SECRET_KEY 'sample secret key'
+ENV DATABASE_URL=postgres://metaci:metaci@postgres:5432/metaci \
+  DJANGO_HASHID_SALT='sample hashid=salt' \
+  DJANGO_SECRET_KEY='sample secret=key' \
+  DJANGO_SETTINGS_MODULE=config.settings.local \
```
davisagli: Should we use $BUILD_ENV here?

jstvz:
Oh, it is also possible to script the selection of the appropriate Chromedriver version. ([Stack overflow](https://stackoverflow.com/a/55266105/733732)). Let me know if you'd rather do that.

Dockerfile
jstvz: This sets `bash` as the default shell for `RUN` instructions. Docker uses `sh` by default, which only checks the exit code of the final command. It also didn't recognize `pipefail` when I tried to set it, so I went with `bash` ([Docs](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#using-pipes)). 

I'm pretty sure that `apt-key` would have a non-zero exit code if there's a problem retrieving the key, but since we may add other RUN instructions containing pipes I think it makes sense to add it now.

davisagli: Got it. Can you add a comment in the Dockerfile with a link to those docs?

davisagli: Never mind, I did.

jstvz:
@davisagli I added a script to download the correct Chromedriver version. I didn't see any references to the `CHROMEDRIVER_VERSION` environment variable, so I also removed it.

PR #1047, opened 2020-04-02
Add meta tags to build detail pages

jstvz:
This commit adds [OpenGraph] and [Twitter Card] `meta` tags. Applications like Quip and Slack will display pages containing these tags with a build status summary that is the same as the email notification, as well as an image indicating the status of the build.

In addition, email templates are updated to remove log lines.

[OpenGraph]: https://ogp.me/
[Twitter Card]: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary


metaci/build/templates/build/detail_layout.html
```diff
@@ -1,4 +1,26 @@
...
+
+{% block base_title %}
+    MetaCI - Build #{{ build.id }}
+{% endblock %}
+
+{% block base_head %}
+    {{ block.super }}
+    <meta name="twitter:card" content="summary" />
+    <meta name="og:type" content="article" />
+    <meta property="og:title" content="MetaCI - Build #{{ build.id }}" />
```
davisagli, line 12: Should we include the repo name in the title?

jstvz: Not sure. Seems like most platforms will truncate `title` to 40-55 characters. It should probably be in the summary at least. I'll add it there.

PR #1096, opened 2020-04-15
Download buildflow assets

davisagli:
Add a way to download BuildFlow assets (currently only the full robot output.xml)

These assets will be linked from the bottom of the build flow tests summary (e.g. https://sfdo-metaci.herokuapp.com/tests/213983/robot)

I've tested it locally.


metaci/testresults/templates/testresults/build_flow_tests.html
```diff
@@ -81,47 +81,59 @@
...
+            <td class="{{ column.status }}">{{ column.value }}</td>
+            {% endfor %}
+          </tr>
+        {% endfor %}
+        <tbody>
+      </table>
+  </div>
+  {% endfor %}
 
+  {% if assets %}
```
davisagli, line 127: The actual change in this file starts here; the rest is just fixing indentation.

metaci/testresults/templates/testresults/build_flow_tests.html
```diff
@@ -81,47 +81,59 @@
...
+  <h3 class="slds-text-heading--small slds-p-top--small">Assets:</h3>
+  <ul class="slds-list_dotted">
+    {% for asset in assets %}
```
boakley, line 130: Will a test always have at least one asset? If not, it might be worth adding some code to add "_No assets available_" or some other message so we know explicitly that there are none, rather than wondering why they aren't showing up.

davisagli: @boakley good idea; added.

PR #1163, opened 2020-05-05
Try to improve on_delete settings

davisagli:
Do more to prevent accidental cascade deletion of builds. Allow cascade deletion of perf summaries.


prescod:
My challenge is knowing when on_delete=models.SET_NULL is safe. How do we know if a view or controller will 500 if a build's org is null. and so forth.

Is there test code that renders the views with some of these values as NULL?

davisagli:
@prescod great point; I'll take another look at that.

davisagli:
@prescod ok, I took a closer look at the fields which are switching to SET_NULL. I'm pretty confident that it's okay to do this for org_instance and current_rebuild because it is already the case that they are not always set.

prescod:
> @prescod ok, I took a closer look at the fields which are switching to SET_NULL. I'm pretty confident that it's okay to do this for org_instance and current_rebuild because it is already the case that they are not always set.

Okay, as long as you've done that review, I'm happy.

PR #1178, opened 2020-05-13
change case enforcement

davisagli:
This lays the groundwork for MetaCI to enforce that releases can only be uploaded if they are linked to a change case.

There is a new URLField on Release called change_case_link. The plan run form has new validation to require selecting a release that has a change_case_link if the selected plan has a role of Release Deploy, Release, or Release Test. The new validation is not in effect by default but can be enabled using the METACI_ENFORCE_RELEASE_CHANGE_CASE setting.

In terms of process this means that release engineers will need to create a Release (in the Django admin) before they can run these plans. The minimum fields to fill out on the Release are repo and (expected) git tag. Once the plan runs and actually creates the tag, a webhook should trigger updating the Release (matching on the repo and git_tag) with additional info.

I made an unrelated change in passing: the plan run form now only shows the "keep org" option if the plan uses a scratch org.


metaci/plan/forms.py
```diff
@@ -86,8 +91,20 @@ def _get_branch_choices(self):
                 choices.append((branch.name, branch.name))
         return tuple(choices)
 
-    def create_build(self):
+    def clean(self):
+        if is_release_plan(self.plan):
+            release = self.cleaned_data.get("release")
+            if not release:
+                self.add_error("release", "Please specify the release.")
```
davidmreed, line 98: Clarifying: the requirement to link a release plan to a release will apply to all MetaCI consumers, but the change case requirement is gated on the environment variable. Right?

davisagli: @davidmreed Yeah, good distinction. That is correct.

PR #1211, opened 2020-05-26
📦 🐈 update dependencies (and fix typescript notices)

davisagli:


src/js/components/perfPages/perfPage.tsx
```diff
@@ -72,7 +72,7 @@ export const UnwrappedPerfPage = ({
   }
 
   // for effects that only run once.
-  const useMount = (f) => useEffect(f, []);
+  const useMount = (f) => useEffect(f, [f]);
```
prescod, line 75: I'm nervous about this one.

https://stackoverflow.com/a/53121021/113477

davisagli: @prescod ah yes, sorry, I just accepted eslint's auto-fix without paying attention to the context. I'll add a comment to ignore the eslint rule for this line instead

davisagli: @prescod https://github.com/SFDO-Tooling/MetaCI/pull/1211/commits/5d737d9afa3cb0dcc22ddf8d5f957b989fc3ec0f -- thanks and good catch

PR #1295, opened 2020-06-26
Remove mock/example/local secrets

davisagli:
Secrets musts be configured, not defaulted. This helps us achieve a clean credential scan.


davisagli:
@boakley thanks. if you're running your copy of MetaCI using Docker you will probably need to add DJANGO_SECRET_KEY and DATABASE_URL to your .env file.

PR #1313, opened 2020-06-30
Feature/robot rest api

boakley:
This adds a new tab on the Repo page that contains links to csv files with robot test results. The csv files are returned by a rest api, and can contain filters for relative dates (this week, last week, etc), along with the repo, branch, test name, source file, and outcome (pass/fail/all). The view doesn't take advantage of some of those filters, but they are there if we need to provide QE with finer grained results. The relative dates allow for the links to be bookmarked. 

Example URL: `http://localhost:8000/api/robot.csv?range=yesterday&repo_name=cci-robot&branch_name=feature/example-1&outcome=Pass`

Two new fields have been added to the TestResults model: 

* `robot_keyword` - the name of the keyword that caused the test to fail
* `robot_tags` - a string which is a comma-separated list of tags for the test

Both the keyword and tags were specifically requested by QE.


metaci/api/views/robot.py
```diff
@@ -0,0 +1,33 @@
...
+    serializer_class = RobotTestResultSerializer
+    renderer_classes = [BrowsableAPIRenderer, JSONRenderer, SimpleCSVRenderer]
+    filterset_class = RobotResultFilter
+
+    def get_queryset(self):
+        queryset = TestResult.objects.filter(method__testclass__test_type="Robot")
+
+        # FIXME: I'd like the user to be able to specify a period
+        # of time such as "last two weeks" or "this month"...
+        # I also need to
```
davisagli, line 25: I think this is probably essential to avoid trying to process too much data in production if the filters are not very selective. I might be inclined to just support specifying an integer number of days if it's easier.

metaci/api/views/robot.py
```diff
@@ -0,0 +1,33 @@
...
+
+        # this is probably not the best default, but it's
+        # good enough for now. Eventually I should add django's
+        # ordering filter.
+        sort = self.request.query_params.get("sort", "id").split(",")
+        queryset = queryset.order_by(*sort)
+
+        return queryset
```
davisagli, line 33: This may be a good place to use https://docs.djangoproject.com/en/3.0/ref/models/querysets/#prefetch-related so that the related objects needed for filtering and serializing are fetched in a single query rather than one per subobject. I suspect that may be a key source of the slow performance you mentioned. Django Debug Toolbar should be useful to check what queries are actually happening.

metaci/api/views/robot.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    A viewset for viewing robot framework test results
+    """
+
+    serializer_class = RobotTestResultSerializer
+    renderer_classes = [BrowsableAPIRenderer, JSONRenderer, SimpleCSVRenderer]
+    filterset_class = RobotResultFilter
+
+    def get_queryset(self):
+        queryset = TestResult.objects.filter(method__testclass__test_type="Robot")
```
davisagli, line 21: We probably need some sort of permission filter so that a user can only see test results for planrepos that they have the "view builds" permission on.

boakley:
> why only expose info about robot test results this way?

Mainly to limit the scope of what I'm doing. I don't have any major heartburn over changing /api/robot to /api/testresults, it's just that it's only returning a small subset of the data specific to my use case and I am concerned that /api/testresults would give the wrong impression. Plus, I don't think anything else will use it in the near future.

I think the better long term solution might be to have the testresults model use generic relationship to either a robot table or an apex table so that we don't have one big monolithic table that has both. I think that's beyond the scope of what I'm trying to do, though.

metaci/api/renderers/csv_renderer.py
```diff
@@ -0,0 +1,42 @@
+# I started here: https://www.django-rest-framework.org/api-guide/renderers/#example
+
+from rest_framework import renderers
+import unicodecsv as csv
```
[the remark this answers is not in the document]
boakley: Makes sense. I think I just blindly copied that code. I'll make that change.

metaci/api/renderers/__init__.py
```diff
@@ -0,0 +1 @@
+from .csv_renderer import SimpleCSVRenderer  # noqa: F401
```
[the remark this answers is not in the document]
boakley: Ok, I'll fix it.

metaci/api/views/robot.py
boakley: ah, `prefetch_related` is something I always have to be reminded about. I make this mistake once every couple of years. Thanks!

metaci/api/tests/test_api_robot.py
```diff
@@ -0,0 +1,149 @@
...
+        self.client.force_authenticate(self.superuser)
+        response = self.client.get("/api/robot/")
+        assert response.status_code == 200
+
+    def xtest_user_access(self):
+        """
+        Make sure the superuser can access the API
+
+        This test fails. Am I doing the test wrong or did
+        I implement the API wrong?
```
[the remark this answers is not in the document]
boakley: The status code is 403 - Forbidden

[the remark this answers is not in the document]
davisagli: I was confused when I looked at this before because the docstring says "make sure the superuser can access the API" but it's actually testing that a NON-superuser can.

DRF views that don't explicitly specify `permission_classes` will use the default permission classes from the settings: https://github.com/SFDO-Tooling/MetaCI/blob/master/config/settings/base.py#L394 -- if this view should be more widely available, it needs to override permission_classes.

metaci/api/views/robot.py
davisagli: @boakley I still don't see anything in here to make sure that the queryset only includes test results for builds that the user has permission to view. Probably something like adding a filter for `build_flow__build__planrepo__in=PlanRepository.objects.for_user(self.request.user)`

PR #1447, opened 2020-08-20
Add test execution errors to each robot test result

boakley:
Execution errors are errors that happen before the test starts, such as errors when importing keyword libraries. This makes sure that those errors appear in the xml we generate for each test.


davisagli:
This is definitely an improvement over the status quo.

However it also makes me wonder: if there were errors prior to running a test, should we short-circuit and abort the entire robot run rather than continuing to attempt running other suites? I could see the argument either way: on the one hand, maybe not all suites are using the same library, and it would be useful to see all the results. On the other hand, setup problems are a pretty clear indication that the build is going to fail and something needs to be fixed, so it would be nice to find that out sooner rather than later.

boakley:
@davisagli You bring up an interesting point. You're right, though, that not all tests use the same library. It's also possible that the failing import is on a library that isn't used at all. 

I'm not sure there's a right or wrong answer here. We can either decide our policy is to let the test run, or our policy can be to stop the test dead in its tracks. Like you, I can see arguments for both. Though to do the latter will require more code since without at least one failing test report, there's no mechanism for reporting the failure (ie: it doesn't show up in the build stdout).

PR #1493, opened 2020-09-09
Week 37 dependency updates

Br4nd0R:


Br4nd0R:
@davisagli - Not sure why `yarn upgrade-interactive` didn't modify the package.json. Should I manually modify, or is there a better solution?

davisagli:
@Br4nd0R I think that's as expected for `yarn upgrade-interactive`: it finds updates that are compatible with the ranges specified in package.json. Add the `--latest` flag to include updates that go beyond those ranges.

davisagli (approved):

PR #1515, opened 2020-09-16
Add features needed for 2gp feature tests

davisagli:
1. Plans can now have an optional commit_status_template field, which can be set to a jinja2 template expression. If it's set, it will be used to calculate the commit status message that will be set in github after a build of the plan completes successfully. (If not set, the old default message will still be used.)

(In the context of 2gp feature tests we will use this to make the plan which builds the package set a commit status that includes the package version id, so that it can be read from there when running the ci_feature_2gp flow.)

2. Plans can now be configured to be triggered by a successful github commit status. The plan regex should be set to match the "context" (name) of the commit status that will be used as a trigger. The github webhook view is now set up to process "status" events in addition to "push" events.

(In the context of 2gp feature tests, we will use this to trigger the plan that runs the ci_feature_2gp flow based on the completion of the plan which uploads the package version.)


Br4nd0R (approved):
This looks good to me!

PR #1608, opened 2020-10-14
optimize deletion; delete test result assets too

davisagli:


metaci/db/management/commands/purge_old_results.py
```diff
@@ -13,22 +14,44 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         year_ago = timezone.now() - timedelta(days=365)
 
+        def show_progress(source, batch_size=1000):
+            i = 0
+            for item in source:
+                i += 1
+                yield item
+                if not i % batch_size:
```
Br4nd0R, line 22: This just outputs until `i` exceeds batch size. Did we want to have anything occur when `len(source) > batch_size` ?

Br4nd0R (approved):

metaci/db/management/commands/purge_old_results.py
davisagli: It outputs every time i is evenly divisible by batch_size.

PR #1719, opened 2020-11-12
[MetaCI] Change default date values for sandbox and production push dates

jofsky:


metaci/build/models.py
```diff
@@ -660,6 +660,14 @@ def run_flow(self, project_config, org_config):
 
         flow_config = project_config.get_flow(self.flow)
 
+        if self.build.plan.role == "release":
+            flow_config["options"].update(
+                {
+                    "sandbox_date": self.build.release.sandbox_push_date,
+                    "production_date": self.build.release.production_push_date,
+                }
+            )
```
davisagli, line 669: @jofsky You're adding to the options for the flow as a whole, but we need to add to the options for one particular step in the flow (github_release_notes). Also, let's be careful not to mutate flow_config, because that might write back to the project config it came from.

I think what we want is:

    if self.build.plan.role == "release" and self.build.release:  # notice I changed the condition
        options = {
            "github_release_notes": {
                "sandbox_date": self.build.release.sandbox_push_date,
                "production_date": self.build.release.production_push_date
            }
        }
    else:
        options = {}

and then add options=options to the args when constructing FlowCoordinator below.

metaci/build/models.py
```diff
@@ -660,6 +660,21 @@ def run_flow(self, project_config, org_config):
...
+        ):
+            for k, v in flow_config.config["steps"].items():
+                if v["task"] == "github_release_notes":
+                    v["options"] = {
+                        "tag": "^^github_release.tag_name",
+                        "publish": True,
+                        "version_id": "^^upload_production.version_id",
+                        "sandbox_date": self.build.release.sandbox_push_date,
+                        "production_date": self.build.release.production_push_date,
+                    }
```
davisagli, line 676: It should not be necessary to do it this way. The FlowCoordinator is supposed to support overriding task options without needing to change the flow config. Also we should only adjust the options that come from the Release object; the others should still come from cumulusci.yml.

I'll take a look at why passing options to the FlowCoordinator wasn't working when I get a chance.

jofsky: Yeah I had to step deep into Pdb to get this to work. Appreciate it @davisagli

metaci/release/models.py
```diff
@@ -46,11 +47,22 @@ class Release(StatusModel):
...
+        blank=True,
+        default=(datetime.date.today() + datetime.timedelta(days=6)).isoformat(),
     )
-    sandbox_push_date = models.DateField(_("sandbox push date"), null=True, blank=True)
     production_push_date = models.DateField(
-        _("production push date"), null=True, blank=True
+        _("production push date"),
+        null=True,
+        blank=True,
+        default=(datetime.date.today() + datetime.timedelta(days=6)).isoformat(),
```
davisagli, line 65: These defaults are not working right because they are calculated when the module is imported, rather than when a Release record is added. You need to set `default` to a function which returns the date, instead of setting a date value directly.

Also, sandbox date should default to the day the release is created instead of the same day as production push, right?

jofsky: @davisagli ah good catch then.

metaci/release/models.py
```diff
@@ -46,11 +47,22 @@ class Release(StatusModel):
...
+        blank=True,
+        default=(datetime.date.today().isoformat()),
     )
-    sandbox_push_date = models.DateField(_("sandbox push date"), null=True, blank=True)
     production_push_date = models.DateField(
-        _("production push date"), null=True, blank=True
+        _("production push date"),
+        null=True,
+        blank=True,
+        default=((datetime.date.today() + datetime.timedelta(days=6)).isoformat()),
```
davisagli, line 65: @jofsky You still haven't fixed these so that evaluating the default is deferred until the object is actually being constructed.

You need to define a function like this:

    def get_today():
        return datetime.date.today()

And then use it in the model like this:

    default=get_today,

Also notice I removed `.isoformat()`. It's a DateField, which means we can give it dates to store. No need to convert to a string and then make it convert back.

You will also need to update the database migrations.

metaci/build/models.py
```diff
@@ -660,6 +660,15 @@ def run_flow(self, project_config, org_config):
 
         flow_config = project_config.get_flow(self.flow)
 
+        # If it's a release build, pass the dates in
+        options = {}
+        if self.build.plan.role == "release" and self.build.release:
+            options["github_release_notes"] = {
+                "sandbox_date": self.build.release.sandbox_push_date,
+                "production_date": self.build.release.production_push_date,
+                "trial_info": self.build.release.trialforce_id,
```
davisagli, line 669:
```suggestion
```

There's no reason to include this one. The trialforce id will never be known when the release is being uploaded, because it needs to be uploaded before it can be installed into the TSO.

metaci/build/tests/test_models.py
```diff
@@ -146,6 +148,27 @@ def test_set_commit_status(self):
...
+                "steps": {1: {"task": "robot_libdoc"}, 2: {"task": "robot_testdoc"}},
+            }
+        }
+        build_flow.run_flow(
+            project_config=project_config,
+            org_config=ScratchOrgInstanceFactory(),
+        )
+        assert build_flow.build.commit_status == "4"
+        build_flow.build.release = Release(repo=RepositoryFactory())
+        breakpoint()
```
davisagli, line 170: @jofsky probably the easiest thing to do would be to make a new method of BuildFlow called _prepare_flow_options and move the logic for building the options there. Then all you need to test it is a BuildFlow that's connected to a Release, and there's no need to actually call run_flow.

metaci/build/models.py
```diff
@@ -686,6 +682,15 @@ def run_flow(self, project_config, org_config):
         # Run the flow
         return self.flow_instance.run(org_config)
 
+    def set_release_dates(self, options):
```
davisagli, line 685: Let's call this `_get_flow_options` instead of `set_release_dates`. It's not setting anything. It's currently only adding the dates, but it would be the natural place to add other flow options. The initial underscore is an indicator that it's not meant to be called from elsewhere in the codebase.

metaci/build/models.py
```diff
@@ -686,6 +682,15 @@ def run_flow(self, project_config, org_config):
...
+        if self.build.plan.role == "release" and self.build.release:
+            options["github_release_notes"] = {
+                "sandbox_date": self.build.release.sandbox_push_date,
+                "production_date": self.build.release.production_push_date,
+                "trial_info": self.build.release.trialforce_id,
```
davisagli, line 690: Still doesn't make sense to include the trial_info (because it isn't known when uploading the release). Please remove it.

davisagli (changes requested):
Real close now.

metaci/fixtures/factories.py
```diff
@@ -46,7 +46,7 @@ def fake_name(prefix=None):
     the class."""
     return factory.LazyAttribute(
         lambda a: (getattr(a, "name_prefix", None) or prefix or "")
-        + factory.Faker("word").generate({})
+        + factory.Faker("word").generate({"locale": None})
```
davisagli, line 49: Adding locale here seems to be causing problems when the tests run on Heroku CI. What was the error you were trying to fix?

jofsky: Couldn't seem tot replicate. Removed it.

davisagli (approved):

PR #1740, opened 2020-11-23
Add "Queued" Column for Builds

Br4nd0R:
* We now show a "Queued" column for builds which displays the date/time at which the build was placed in the queue.
These changes apply to the following templates:
     * `build/templates/build_list.html`
     * `build/templates/detail_rebuilds.html`
     * `plan/templates/detail.html`
     * `plan/templates/plan_repo_detail.html`
     * `repository/templates/repo_detail.html`

* Note: The "Commit" column has been removed from `build/templates/build_list.html` as it seemed the least useful and I wanted to reduce clutter with the addition of the new date/time field.

* The `build` model now has a `get_commit` method which returns a truncated version of the commit SHA. This will help reduce clutter when displaying commits in build info tables. The full commit SHA is still visible on the `build/detail_layout.html` template.

* `BuildFactory` has been updated to generate a commit(ish) string for test builds.

### DIFF
--- a/metaci/build/models.py
+++ b/metaci/build/models.py
@@ -262,6 +262,10 @@ class Build(models.Model):
     def get_error_message(self):
         return self.get_build_attr("error_message")
 
+    def get_commit(self):
+        """Return the first 6 characters of the commit sha, followed by an elipsis"""
+        return f"{self.commit[:6]}..."
+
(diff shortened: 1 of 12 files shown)


davisagli:
Closing in favor of #1741

PR #1741, opened 2020-11-23
Show enqueue time if build not started

### LINKED ISSUE #961: Add Build "Queued Date" Build Lists
Currently, "Start Date" and "End Date" are visible in various build lists which is usually useful but does show exactly when a build was Queued which can be useful information especially when builds get backed up.

Screenshot 2020-02-28 16 46 47

This is especially helpful when a build is Queued and therefore doesn't have a Start or End date displayed.

Screenshot 2020-02-28 16 47 23

Br4nd0R:
* We now display the date/time that a build entered the queue if it has not yet started running.  The queued date/time will be denoted by the string "Queued" before the date/time in the "Start Time" column. For example: `Queued Aug. 14, 2019, 6:19 p.m.`. This change has been made to the following templates:

    * `build/build_list.html`
    * `cumulusci/org_detail.html`
    * `cumulusci/org_instance_detail.html`
    * `plan/detail.html`
    * `plan/plan_repo_detail.html`
    * `repository/branch_detail.html`
    * `repository/commit_detail.html`
    * `repository/repository.html`

* If the build has already started or finished then this column will display the actual start date/time (this is the current behavior).

* There is a new template `build/build_table.html` that is now used to show the table of build information on other tempaltes.

* To save on column space we now show a truncated version of the commit SHA (the first eight chars followed by an ellipsis). We still show the full commit SHA in templates: `branch_detail.html` and `commit_detail.html` 

* Closes #961


metaci/build/templates/build/build_list.html
```diff
@@ -60,12 +60,18 @@
         <div class="slds-truncate" title="{{ build.branch.name }}">{{ build.branch.name }}</div>
       </td>
       <td data-label="Commit">
-        <div class="slds-truncate" title="{{ build.commit }}">{{ build.commit }}</div>
+        <div class="slds-truncate" title="{{ build.get_commit }}">{{ build.get_commit }}</div>
```
davisagli, line 63: We can remove the `title` attribute (which controls the tooltip) in the cases where it's the same thing that is shown in the actual table cell.

Do you think it would make sense to make all these build listings include a common template for the table so we don't have to make these changes in so many places?

metaci/build/models.py
```diff
@@ -283,6 +283,9 @@ def get_time_qa_start(self):
     def get_time_qa_end(self):
         return self.get_build_attr("time_qa_end")
 
+    def get_commit(self):
+        return f"{self.commit[:8]}..."
```
davisagli, line 287: Let's leave off the ellipsis. It'll make it easier to copy, and using an abbreviated git commit hash is a common convention.

metaci/build/templates/build/build_list.html
Br4nd0R: Yes, a `build_table.html` template would be a nice thing right now :D

davisagli: I'll leave it up to you whether to do that now or in a separate PR.

Br4nd0R:
@davisagli - New `build_table.html` template included to show build info on other templates.

davisagli (approved):

PR #1742, opened 2020-11-24
Header Tweaks

Br4nd0R:
* Removed values for `parent_link` in templates where this was linking to `/builds`, `/plans`, or `/repos` views. These views are always available in the nav bar via the "Builds", "Plans", and "Repos" links.

* Headers titles have been adjusted to show text that includes the records type with the larger text below for id/name of the record. Rebuild headers were further tweaked so that Build and Rebuild numbers display in a horizontal fashion.

Screen Shot 2020-11-24 at 2 24 36 PM
Screen Shot 2020-11-24 at 2 25 16 PM

* Places where raw url text was serving as a hyperlink have been replaced with buttons for viewing repositories, branches, and commits, in GitHub when viewing the respective detail pages. If the URL was providing additional context (like the particular Repository) that information has been included in the detail view. See below screenshot for button in top right for navigating to the specific commit, along with a link to the repository record in MetaCI:
Screen Shot 2020-11-24 at 2 26 26 PM


metaci/build/templates/build/build_list.html
```diff
@@ -1,7 +1,5 @@
 {% extends 'layout_full.html' %}
 
-{% block layout_parent_link %}<a href="/builds">Builds</a>{% endblock %}
```
Br4nd0R, line 3: This link is redundant. "Builds" is available (and more prominent) in main navbar

davisagli (approved):
NIce tweaks

PR #1743, opened 2020-11-25
Improved GitHub commit status

Br4nd0R:
* Builds that error will now have the following commit message: 💥 An error occurred during the build
* Builds that fail will now have the following commit message: ⚠ ️ 20/980 failed
* Failed builds in GitHub will now have the "details" link to the failed tests instead of the build detail page layout.


Br4nd0R:
I'm unsure of the best way to test `metacirepository.utils.create_status`:
* If I attempt to make the build a mock object, then I end up needing to recreate getter methods (like get_status) on the mock object. Seems like a lot of work to get the mock object to work properly.
* If I attempt to use `BuildFactory`, the given build object has a bad value for `build.repo.github_api`. I'm also unable to set the attribute on the build object to `mock.Mock()`.
* Should I mock out all the necessary `github3` responses and attempt to mock nothing at all?

Br4nd0R:
@davisagli - I believe the code here is ready to go. The only other reference to `.github_api` I see in the codebase is in`metaci/repository/tasks.py` and is on a `Branch` object, not a `Repository`.

I still need to figure out how to QA this in GitHub with my local.

metaci/repository/tests/test_utils.py
```diff
@@ -0,0 +1,191 @@
...
+    build.plan.context = mock.Mock()
+    build.repo.get_github_api = mock.Mock(return_value=repo)
+    return build, repo
+
+
+@pytest.mark.django_db
+def test_create_status__queued():
+    build, repo = setup_build_with_status("queued")
+
+    with mock.patch("metaci.build.models.settings.SITE_URL", "initech.co"):
```
davisagli, line 27: Instead of doing this separately for each test, you can make a fixture that is automatically initialized once before all tests in the module, and torn down after they're all done:

    @pytest.fixture(scope="module", autouse=True)
    def site_url():
        with mock.patch("metaci.build.models.settings.SITE_URL", "initech.co"):
            yield

(see https://docs.pytest.org/en/stable/fixture.html#autouse-fixtures-xunit-setup-on-steroids)

metaci/repository/utils.py
```diff
@@ -1,48 +1,63 @@
...
+    target_url = None
+    description = None
+    repo = build.repo.get_github_api()
 
     if build.get_status() == "queued":
         state = "pending"
         description = "The build is queued"
-    if build.get_status() == "waiting":
+
+    elif build.get_status() == "waiting":
```
davisagli, line 18: This is a tiny thing and it was already there, but let's do `status = build.get_status()` once and then use it so we aren't repeatedly calling get_status()

metaci/repository/utils.py
```diff
@@ -1,48 +1,63 @@
...
         state = "failure"
         if build.plan.role == "qa":
-            description = "{} rejected. See details for QA comments".format(
-                build.qa_user
-            )
+            description = f"{build.qa_user} rejected. See details for QA comments"
         else:
-            description = "Tests failed"
+            plural = build.tests_fail != 1
+            description = f"⚠ ️There {'were' if plural else 'was'} {build.tests_fail} test{'s' if plural else ''} that failed."
```
davisagli, line 49: github doesn't leave much space for the commit status message. I was thinking of something like "⚠ xx/yy tests failed" where xx is the # of failed tests and yy is the total number of tests.

metaci/repository/utils.py
```diff
@@ -1,48 +1,63 @@
...
+    else:
+        raise BuildError(
+            f"ERROR: Unrecognized build status encountered: {build.get_status()}"
+        )
 
     response = repo.create_status(
         sha=build.commit,
         state=state,
-        target_url=build.get_external_url(),
+        target_url=target_url or build.get_external_url(),
```
davisagli, line 60: Initialize `target_url = build.get_external_url()` on line 10, and then this line can simply use `target_url`.

davisagli (changes requested):

davisagli (approved):
LGTM, but I think we'll have to wait for the Heroku incident to be resolved in order to run the tests.

davisagli (approved):

PR #1764, opened 2020-11-26
f strings!

Br4nd0R:
All but 6 references to `str.format()` now use f strings. One of these references I believe should remain (under `metaci/release.py` in `METACI_CHANGE_CASE_URL_TEMPLATE`). The other references are questionable.

For example, there are several in `robot_importer.py` that use an odd double quote syntax like: `'"{}"'.format(screenshot)`. I'm not sure if the double quotes are there for a reason, or if this was just a mistake that is working by chance.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -246,7 +246,7 @@ ROOT_URLCONF = "config.urls"
 
 # Location of root django.contrib.admin URL, use {% url 'admin:index' %}
 ADMIN_URL = env("DJANGO_ADMIN_URL", default="admin")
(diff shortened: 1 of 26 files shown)


davisagli:
> For example, there are several in robot_importer.py that use an odd double quote syntax like: '"{}"'.format(screenshot). I'm not sure if the double quotes are there for a reason, or if this was just a mistake that is working by chance.

It's just adding double-quotes around the screenshot path. It can be replaced with `f'"{screenshot}"'`

Btw, looks like this needs to have some conflicts resolved.

davisagli (approved):

PR #1771, opened 2020-11-30
Improve layout for tags

Br4nd0R:
* We now change the header to properly display the "Tag" title when a branch is referencing a GitHub tag.
* The button in the header bar now displays with "View Tag in GitHub" if the branch references a tag.


metaci/repository/templates/repository/branch_detail.html
```diff
@@ -3,20 +3,38 @@
...
-    <li>
-      <a href="{{ repo.url }}/tree/{{ branch.name }}" target="_blank">
-        <button class="slds-button slds-button_outline-brand">View Branch in GitHub</button>
-      </a>
-    </li>
-  </ul>
+    <ul class="slds-button-group-list">
+      <li>
+        {% if branch.is_tag %}
+          <a href="{{ repo.url }}/releases/tag/{{ branch.name }}" target="_blank">
```
davisagli, line 24: Need to cut of "tag: " here, right?

davisagli (approved):

PR #1785, opened 2020-12-06
use register decorator

Br4nd0R:
* Use the `@admin.register` decorator
* Black also tidied up a couple things


davisagli (approved):
Thanks!

Could you do a separate PR to run black on the entire metaci folder so we can make future changes without mixing in formatting changes?

PR #1786, opened 2020-12-07
Fix Robot Test Result Importing

Br4nd0R:
* Fixed an issue where executing multiple robot tasks in a single flow caused only the last robot tasks' results to be created.
    * After each task executes we check if the task is a robot task (by looking at `StepSpec.task_class`) and generate related test result records if it is.
* Removed django `TestCase` class in `test_robot_importer` in favor of `@pytest.mark.django_db` on individual methods.
* Improved tests to ensure that the correct assets are generated for output.xml and screenshot files.


metaci/build/models.py
```diff
@@ -701,21 +700,19 @@ def record_result(self):
...
             # FIXME: Not sure why robot stopped writing into the cwd
             # (build temp dir) but this should handle it so long as
             # only one build runs at a time
             has_results = True
-            try:
-                import_robot_test_results(self, root_dir_robot_path)
-            finally:
-                os.remove(root_dir_robot_path)
+            # import_robot_test_results(self, root_dir_robot_path)
+            root_dir_robot_path.unlink()
```
davisagli, line 715: I believe this whole section (lines 703-715) can go away.

Removing output.xml should be done in the same place where the robot results are imported (where we know for a fact where it is).

As we discussed the other day, we don't need `has_results` at all. We can simply remove the condition on line 749 and always do those queries to count the results.

Br4nd0R:
Got thinking about this more over the weekend, and now I'm wondering if there should be some sort of integration test to run a flow that contains two robot tasks and assert that each of the robot tasks has the proper results associated with it.

Or is it overkill, because we know that `import_robot_test_results()` does create test results for a single `FlowTask` (what current tests validate).

prescod:
```
        if step.task_class == Robot:
            import_robot_test_results(flowtask, "output.xml")
```

I'm not a fan of this idiom. The minimum required to break this is that we subclass the Robot class. But even using isinstance is only a slight improvement in this case.

The right fix in these kinds of cases is usually to use some kind of polymorphism.

```
        if step.robot_format_results:
            import_robot_test_results(flowtask, step.robot_format_results)
```

Theoretically we could try and go a step further and ask the task to parse its own results so that the format on the filesystem is irrelevant and any testing library can inject its results into MetaCI, but maybe we shouldn't let the best be the enemy of the good.

isinstance (not ==) is useful for when polymorphism fails because you can't inject methods into the other class. In this case we control both classes so we can.

(and as an aside, one reason the Julia language is gaining popularity is because you can handle the cases even when you don't control the other class, if you ever have 35 minutes to share, I enjoyed this talk https://www.youtube.com/watch?v=kc9HwsxE1OY )

davisagli:
@prescod I agree in principle with what you're saying, and like the idea of moving responsibility for parsing test results into a common format to a (optional) method on the task class. However, the present goal is to do the simplest thing possible to support multiple robot tasks in the same flow, because there are teams that will use it as soon as they can. So I'd rather not expand the scope to include that refactor. We aren't doing anything that will preclude refactoring in that way later.

prescod:
@davisagli moving the code is probably a lot of work.

Using polymorphism instead of task==Robot to find out whether "output.xml" should be available is very little work (although it does span two different projects).

davisagli:
@prescod But I'd like to wait to do that until we actually have two different tasks that we want to get robot results from. Otherwise I think we're prone to invent an interface that doesn't actually cover everything we need it to.

prescod:
Okay I guess we'll have to agree to disagree but it isn't an urgent issue in this case.

metaci/build/flows.py
```diff
@@ -9,24 +12,34 @@ class MetaCIFlowCallback(FlowCallback):
...
-        flowtask = FlowTask.objects.find_task(self.buildflow_id, step.path, step.step_num)
+        flowtask = FlowTask.objects.find_task(
+            self.buildflow_id, step.path, step.step_num
+        )
         flowtask.time_end = timezone.now()
         flowtask.result = result.result
         flowtask.return_values = result.return_values
+
+        if step.task_class == Robot:
+            import_robot_test_results(flowtask, "output.xml")
```
davisagli, line 37: Part of the goal is to automatically find output.xml in the location where the robot task put it, and no longer require customizing the task in MetaCI to put it in the root folder of the repository.

Probably the cleanest way to do this is:
- Adjust the Robot task in cumulusci to set self.return_values["robot_outputdir"] to the outputdir path that was passed to robot itself.
- Then this code can do something like `if "robot_outputdir" in result.return_values:`

(This is a variation on what @prescod was suggesting, but indicating that the task has robot results in return_values instead of providing a method to access the results. I think the former will work better since we don't have the task instance anymore in this callback.)

davisagli:
@Br4nd0R Regarding the question about tests...it would be good to have a test that calls post_task twice with different results, and confirms that the expected db records were created. It's probably best if this uses robot results that were generated ahead of time and stored as part of the tests, so that we don't actually need to run Robot Framework in order to test MetaCI.

metaci/build/flows.py
Br4nd0R: We now [set outputdir on the return_values](https://github.com/SFDO-Tooling/CumulusCI/pull/2252) for the Robot task.

prescod: Seems like a great solution to me.

Br4nd0R:
@davisagli - Feel free to look at this in the morning. I'm attempting to test the execution of `post_task()` twice for two different instances of Robot tasks. On my latest commit I'm getting an odd error on the following:
```
 pytest -l --disable-warnings metaci/build/tests/test_flows.py
```
The error is complaining that the the `flowtask` that's being related to a `TestResult` needs to be saved. The code in `django.db.models.base.py` makes it seem like the `flowtask` doesn't have a value for its primary key, yet I have included `save()` calls for each of the `flowtasks` that I create in the test.

davisagli:
@Br4nd0R The first line of post_task calls FlowTask.objects.find_task, which looks for a task with matching build_flow_id, path, and step_num. If it doesn't find one, it constructs a new FlowTask, but does not save it.

It's normal for the FlowTask to not exist yet. This is how it gets created. But it needs to be saved before other models can be related to it. I think you simply need to move the import_robot_test_results call below flowtask.save() in post_task.

By the way, in the test I'd suggest using real StepSpec and StepResult objects instead of mocks. They are pretty straightforward objects to construct and it'll make the test more realistic.

metaci/testresults/robot_importer.py
```diff
@@ -1,85 +1,99 @@
...
+            else:
+                screenshot_path = Path(screenshot)
+
             with open(screenshot_path, "rb") as f:
                 asset = BuildFlowAsset(
-                    build_flow=build_flow,
+                    build_flow=flowtask.build_flow,
                     asset=ContentFile(f.read(), screenshot),
-                    category="robot-screenshot",
+                    category=f"robot-screenshot-{i+1}",
```
davisagli, line 77: They can and should all go in the same category. It's the name passed to ContentFile that needs to be unique. We should also incorporate the step_num into that name.

davisagli (changes requested):

metaci/build/tests/test_flows.py
```diff
@@ -67,70 +72,135 @@ def test_post_task__result_has_exception(get_spec):
...
-        path="Robot",
-        result="something",
-        return_values={"robot_outputdir": str(path / "robot_1.xml")},
-        exception=None,
-    )
+def test_post_task__single_robot_task(get_spec):
+    with temporary_dir() as output_dir:
+        output_dir = Path(output_dir)
+        open((output_dir / "selenium-screenshot-1.png"), mode="w+")
+        open((output_dir / "selenium-screenshot-2.png"), mode="w+")
```
davisagli, line 79: fwiw you can use the `touch` function from cumulusci.utils to create a file.

metaci/build/tests/test_flows.py
```diff
@@ -0,0 +1,209 @@
...
+        tr_method = TestMethod.objects.get(name="Via UI")
+        test_ui = TestResult.objects.get(method=tr_method)
+        assert 1 == test_ui.assets.count()
+
+        # Three tests total between the two output files
+        assert 3 == TestMethod.objects.all().count()
+
+
+def copy_file(input_file, output_file):
+    """Copy contents of input_file to output_file"""
```
davisagli, line 202: You can use shutil.copyfile for this: https://docs.python.org/3/library/shutil.html#shutil.copyfile

davisagli:
I tested this as follows:
1. Made a feature/robot-screenshot branch of CumulusCI-Test which always captures a screenshot during the robot test
2. On my local instance, set up a Feature Test - Robot plan which runs the robot task twice with different output dirs.
3. Ran the plan and verified the results.
I had to adjust the cumulusci version in requirements.txt to make sure I have the prerelease where it sets robot_outputdir, but with that in place, it's looking good. It created 2 robot-output assets, twice the normal number of tests, and each UI test successfully links to a different screenshot file.

I made a couple tiny suggestions above but this is looking really close.

Br4nd0R:
@davisagli - Thanks for the review. I've gone ahead and included `cumulusci.utils.touch` and `shutil.copyfile`.

davisagli (approved):

PR #1813, opened 2020-12-12
Order releases by date created rather than production push date (which is optional)

jofsky:


metaci/release/models.py
```diff
@@ -85,7 +85,7 @@ def get_production_date():
 
     class Meta:
         get_latest_by = "production_push_date"
-        ordering = ["-production_push_date"]
+        ordering = ["-git_tag"]
```
davisagli, line 88: This won't work right -- the sorting for a text field is alphabetical, so a tag like release/9 will sort before release/10 in descending order.

If the goal is to show the most recently created release first, I'd suggest `-created`

davisagli (changes requested):

metaci/release/models.py
davisagli: You'll also need to add a migration.

davisagli:
@jofsky whenever you change a model, you need to generate a new migration so that Django can keep track of whether other instances of the project need to have their databases updated.

Run `./manage.py makemigrations -n NAME` where NAME is a brief description of what changed (e.g. `release_ordering` in this case). Then run black on the resulting migration file and commit it.

davisagli:
Django docs on migrations, for more info: https://docs.djangoproject.com/en/3.1/topics/migrations/

jofsky:
Ah shoot, I forgot to commit it before requesting. Although I will read the docs. @davisagli

davisagli (approved):

PR #1829, opened 2020-12-17
Replace django-rq-scheduler

davisagli:
This switches us from using django-rq-scheduler (which is not maintained) and db models for configuring scheduled jobs, to scheduling them based on a Django setting.

- Removed django-rq-scheduler from requirements and INSTALLED_APPS. Its tables remain in the database; we can add a migration to remove them later once we're sure they're no longer used.
- Turned old migrations which created RepeatableJob records into no-ops, and removed the metaci_scheduled_jobs management command.
- Added a metaci_rqscheduler management command which wraps the rqscheduler command from django-rq. This ensures that scheduled jobs are up to date in redis based on what's configured in the CRON_JOBS setting. This is done when the command starts before running the scheduler.

Side effects:
- Changing the schedule will now require a release.

Other options I considered for where to update the scheduled jobs:
- In a Django app ready callback (rejected because it would run once for every dyno, whereas the rqscheduler command only runs in one dyno)
- release phase (rejected because it would not update the scheduled jobs when running locally)
- Django checks framework (rejected because it would run for some commands that are unrelated to scheduling jobs)

My biggest open question is: does it make sense to allow for configuring the jobs via a generic CRON_JOBS setting like this? The CRON_JOBS setting makes it possible to override the default jobs to disable one of them or change the schedule, but perhaps makes it too easy to screw things up. Instead it might make sense to hardcode a lot of the job parameters and only expose specific ones that we want to make configurable (e.g. what time to run the daily builds).


Br4nd0R:
Responding to your open question: I think that more limited set of configuration options in exchange for something that is more difficult to break is worth it.

prescod:
I didn't see any problems but it is admittedly tricky to keep the various abstraction layers in mind. Redis, Redis Queue (RQ), RQScheduler, Django-RQ, MetaCIRQScheduler. (and then my own build scheduler on top again)

But yeah, you've eliminated the layer that allowed manipulating jobs through the Django Admin and you use a hard-coded JSON (supplied as Env or hard-coded Python) instead.

One middle ground between having to replace everything and only being able to configure selected aspects would be a JSON merge of the env-var and the default config similar to a CCI merge.

jofsky:
@davisagli is there a way we can just set defaults rather than hardcode the values or have scheduled jobs to reset the defaults to whatever is checked into source? I'm just wondering if there's any value in having admins do one off builds, but with that said if it means disabling or overriding default jobs I wonder if this is a safe suggestion.

davisagli:
I adjusted this so that it's possible to configure the cron schedule for specific jobs (or disable them) without replacing the entire default configuration.

Br4nd0R (approved):
Nice change 👍 , I should go re-review the django3 upgrade stuff again.

PR #1887, opened 2021-01-07
Django 3 Upgrade

Br4nd0R:
Updated MetaCI to Django to 3.1.5

# Changes to get tests passing
* Removed the use of the `python_2_unicode_compatible` decorator
* Switched from `staticfiles` to `static`. The prior was [deprecated in Django 2.1 ](https://docs.djangoproject.com/en/2.2/releases/2.1/#features-deprecated-in-2-1)
* Switched from `django.conf.urls.url()` to `django.conf.ursl.re_path()`. The prior was [deprecated in Django 3.1](https://docs.djangoproject.com/en/3.1/ref/urls/#url)

# Configuration Changes/Considerations
* Django 3 dropped support for PostgreSQL 9.4. MetaCI currently shows `PG Version: 12.4` (`heroku pg:info -a sfdo-metaci`)
* We now use `gettext_lazy()` instead of `ugettext_lazy()`

# Stil Outstanding
- [x] `django.db.backends.postgresql_psycopg2` module has been removed and we reference it the `docker_ec2.rst` file and also have `pscopg2` as a dependency in `base.txt` but not sure if this is an issue or not.
- [x] Should we set [LANGUAGE_COOKIE_HTTPONLY](https://docs.djangoproject.com/en/3.1/ref/settings/#language-cookie-httponly) to `True` (currently defaults to `False`)? We currently set `SESSION_COOKIE_HTTPONLY = True`, not sure if we also need to do this for the language cookie... my guess is no.

## Changes for fewer pytest warnings
These aren't directly related to the upgrade, but were some low-hanging warnings that I was able to get rid of when running pytest.
* Included a `__test__ = False` on several classes so that Pytest doesn't complain about them.
* Documented `providing_args` for the `build_complete` signal in a comment as [they are for documentation only](https://docs.djangoproject.com/en/3.0/topics/signals/#defining-signals).
* Switched a call from `logger.warn` to `logger.warning` (the prior is deprecated)
* Removed the use of `assertDictContainsSubset` as it is deprecated

## QA Checks Performed

- [x] Ran a feature test plan against `main` and `feature/brandon-test` in the CumulusCI-Test repo

Initial runs of jobs gave `objc[4506]: +[NSValue initialize] may have been in progress in another thread when fork() was called.`. Adding `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` fixed this issue. I added it to the `env.example` file.

- [x] Ran a plan which runs two robot tasks in a single flow (just for fun).

Verified that jobs completed successfully, and that test results/methods/screenshots look correct.


davisagli:
> django.db.backends.postgresql_psycopg2 module has been removed and we reference it the docker_ec2.rst file and also have pscopg2 as a dependency in base.txt but not sure if this is an issue or not.

We still need psycopg2 as a dependency in order for Django's built-in postgresql support to work (see https://docs.djangoproject.com/en/3.1/ref/databases/#postgresql-notes) but it is now implemented in django.db.backends.postgresql rather than the removed django.db.backends.postgresql_psycopg2

davisagli:
> Adding OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES fixed this issue. I added it to the env.example file.

I've run into this too; this seems like a good approach.

config/urls.py
```diff
@@ -14,62 +13,62 @@
...
-    url(
+    re_path(r"^$", build_views.build_list, name="home"),
+    re_path(
         r"^about/$",
         mbci_views.AboutView.as_view(template_name="pages/about.html"),
         name="about",
     ),
-    # Django Admin, use {% url 'admin:index' %}
-    url(settings.ADMIN_URL_ROUTE, admin.site.urls),
+    # Django Admin, use {%  re_path 'admin:index' %}
```
davisagli, line 26:
```suggestion
    # Django Admin, use {%  url 'admin:index' %}
```

I think this templatetag example was accidentally included in your search-and-replace

config/urls.py
```diff
@@ -14,62 +13,62 @@
...
+    re_path(r"^webhook/github/push$", github_webhook, name="github_webhook"),
 ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
 # override default error view
 handler403 = "metaci.views.custom_403"
 
 if settings.DEBUG:
     # This allows the error pages to be debugged during development, just visit
-    # these url in browser to see how these error pages look like.
+    # these  re_path in browser to see how these error pages look like.
```
davisagli, line 52:
```suggestion
    # these paths in the browser to see how these error pages look.
```

metaci/users/models.py
```diff
@@ -1,15 +1,13 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals, absolute_import
+from __future__ import absolute_import, unicode_literals
```
davisagli, line 2: We can remove these; they are the default in Python 3.

davisagli (commented):
Nice work so far

davisagli:
> We currently set SESSION_COOKIE_HTTPONLY = True, not sure if we also need to do this for the language cookie... my guess is no.

It should be set to true for the language cookie unless the cookie needs to be accessed by client-side scripts, which it doesn't.

davisagli (approved):
This looks good to me.

PR #1888, opened 2021-01-07
Remove reference to metaci cli

Br4nd0R:
* This removes the references to MetaCI cli in readme.rst. 

Outstanding item:
* I've replaced the mention of the cli with a note that MetaCI can be run locally or deployed into a production environment on Heroku. For local development we can link to `running.rst`. We have the "deploy to heroku" button, but do we have anything to point people to for setup on Heroku or is this a gap in our docs?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,11 +54,9 @@ You can also fork the CumulusCI-Test repository and use that as a demo since it 
 Getting Started
 ---------------
 MetaCI can be run locally or configured for use on Heroku.
-For local setup see `running`_
+For local setup see :ref: `running`.
 
-You can also reference the following session recordings showing live demos of how to configure `CumulusCI` and `MetaCI` for projects:
-https://www.youtube.com/watch?v=CgvYKDqb6Ng
-https://www.youtube.com/watch?v=TN-e785MqBA
+You can also reference `this session <https://www.youtube.com/watch?v=TN-e785MqBA>`_ from TrailheadDX '18 showing a live demo of how to configure `CumulusCI` and `MetaCI` for projects.
 
 Or, if you feel like deploying manually and configuring through the Admin interface:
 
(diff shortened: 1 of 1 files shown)


davisagli:
@Br4nd0R We probably need to verify that the "deploy to heroku" button still works, and write some docs about that. We have a doc that we can share internally, but it has a lot of things that wouldn't necessarily be requirements for instances run by 3rd parties.

README.rst
```diff
@@ -53,35 +53,16 @@ You can also fork the CumulusCI-Test repository and use that as a demo since it
...
 
-The recommended way to deploy and configure `MetaCI` is via the `MetaCI-CLI` project which can access your local CumulusCI project configurations including repo info and org configs to quickly get you up and running with `MetaCI`.
-
-Full documentation on launching and configuring `MetaCI` on Heroku using the `MetaCI-CLI`:
-http://metaci-cli.readthedocs.io/
-
-You can also reference the following session recordings showing live demos of how to configure `CumulusCI` and `MetaCI` for projects:
-https://www.youtube.com/watch?v=CgvYKDqb6Ng
-https://www.youtube.com/watch?v=TN-e785MqBA
+You can also reference `this session <https://www.youtube.com/watch?v=TN-e785MqBA>`_ from TrailheadDX '18 showing a live demo of how to configure `CumulusCI` and `MetaCI` for projects.
```
davisagli, line 59: This video uses the MetaCI-CLI commands that no longer work, so we probably should recommend it, at least not for the specific setup process.

Br4nd0R:
@davisagli - Could we remove the button for now and make a new story for writing additional docs on deploying to Heroku? I'd rather get rid of the pieces that could cause (more) confusion and lead to us fielding additional questions or leading people down paths that are no longer viable.

README.rst
Br4nd0R: Oh, I know the first video did, but I didn't see any use of the MetaCI cli when I scanned this video. I can remove it too.

davisagli: I clicked around a bit and saw it at 24:58

davisagli (approved):
Thanks, I've been meaning to do this for a long time.

PR #1905, opened 2021-01-19
add 'lang' attribute to <html> tag

Br4nd0R:


prescod:
I don't know the history of django_slds so I won't approve but it looks fine to me.

Br4nd0R:
> I don't know the history of django_slds so I won't approve but it looks fine to me.

Nor do I 😬

davisagli (approved):

davisagli:
django-slds is a package that was created here, before my time, with base Django templates that use SLDS CSS classes. https://github.com/SalesforceFoundation/django-slds

PR #1979, opened 2021-02-10
Blank regex fix

Br4nd0R:
# Changes
* MetaCI now skips creating builds for plans that it detects have a non-manual trigger type _and_ no regex specified. We now log when this even occurs.
* We now throw a validation error that can be seen in the Admin UI for this same scenario.


metaci/repository/views.py
```diff
@@ -216,6 +219,11 @@ def create_builds(event, payload, repo, branch, release):
     for pr in repo.planrepository_set.should_run().filter(
         plan__trigger__in=["commit", "tag", "status"]
     ):
+        if not pr.plan.regex:
+            logger.warn(
```
davisagli, line 223:
```suggestion
            logger.warning(
```

(warn works but is deprecated)

davisagli (changes requested):

davisagli (commented):
Looks great. I didn't actually QA it -- should I?

Br4nd0R:
> Looks great. I didn't actually QA it -- should I?

I've QA'ed the Admin UI portion already. Per our conversation in slack, I can QA that plans get skipped and logging is output in staging.

davisagli (approved):
Merge when you're ready

PR #2031, opened 2021-02-25
[MetaCI] Gus Bus Implementation Steps

jofsky:


metaci/release/models.py
```diff
@@ -17,6 +17,28 @@ def __str__(self):
         return self.name
 
 
+class ImplementationStep(models.Model):
+    description = models.ForeignKey(
+        "plan.Plan", on_delete=models.CASCADE, related_name="name"
+    )
```
davisagli, line 23: We should probably leave this out and just use step.plan.name when we need a name/description. It doesn't make sense to have a field named description which is actually storing a lookup to a Plan model

metaci/release/models.py
```diff
@@ -17,6 +17,28 @@ def __str__(self):
...
+    plan = models.ForeignKey(
+        "plan.Plan", on_delete=models.CASCADE, related_name="plans"
+    )
+    start_time = models.DateField(
+        _("start_time"),
+        null=True,
+        blank=True,
+        default=None,
+    )
+    stop_time = models.DateField(_("stop_time"), null=True, blank=True, default=None)
```
davisagli, line 36: These should be DateTimeField so we can also set a time, and they should not have null=True or blank=True (implementation steps need to be planned for a specific window of time)

metaci/release/migrations/0014_implementationstep_implementation_step_id.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("release", "0013_implementationstep"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="implementationstep",
+            name="implementation_step_id",
```
davisagli, line 15: How is it that you have a migration adding this field but it's not in models.py? Maybe call it "external_id" intead of "implementation_step_id" to avoid repeating the model name and to distinguish it from the primary key in the django db.

jofsky: Ah apologies my code isn't pushed @davisagli

metaci/release/models.py
```diff
@@ -79,6 +102,12 @@ def get_production_date():
     change_case_link = models.URLField(
         _("change case link"), max_length=1024, null=True, blank=True
     )
+    sandbox_implemention_step = models.ForeignKey(
+        "release.ImplementationStep",
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="releases",
+    )
```
davisagli, line 110: It doesn't make sense to have a release field on ImplementationStep _and_ implementation step fields on Release. One direction is enough to create a relationship; 2 directions will be difficult to keep in sync.

I don't think we need the field on Release, because we can tell which ImplementationStep is the sandbox push one based on which Plan it's related to.

metaci/release/models.py
```diff
@@ -17,6 +17,29 @@ def __str__(self):
         return self.name
 
 
+class ImplementationStep(models.Model):
+    release = models.ForeignKey(
+        "release.Release", on_delete=models.CASCADE, related_name="releases"
```
davisagli, line 22: related_name should be "implementation_steps"

Then it will be possible to get a queryset for a release object's implementation steps using release.implementation_steps

related_name is the name that's used on the _other_ object.

metaci/release/models.py
```diff
@@ -17,6 +17,29 @@ def __str__(self):
...
+    )
+    plan = models.ForeignKey(
+        "plan.Plan", on_delete=models.CASCADE, related_name="plans"
```
davisagli, line 25:
```suggestion
        "plan.Plan", on_delete=models.CASCADE, related_name="implementation_steps"
```

metaci/build/models.py
```diff
@@ -321,7 +326,21 @@ def run(self):
...
+                        f"Error while sending implementation stop step webhook: {err}"
+                    )
+                    self.logger.error(message)
+                    set_build_info(
+                        build,
+                        status="error",
+                        exception=message,
+                        time_end=timezone.now(),
+                    )
+                    return
```
davisagli, line 343: @jofsky I don't think we need explicit error handling for this (lines 332-343); it should get caught by the exception handler on line 332.

metaci/build/models.py
```diff
@@ -321,7 +326,21 @@ def run(self):
 
             # Initialize the project config
             project_config = self.get_project_config()
-
+            if self.plan.role == "push_sandbox" or self.plan.role == "push_production":
```
davisagli, line 329: Can we put the logic for determining whether to call CTC into its own method (maybe `should_send_start_stop`) -- then we can use that both here and on line 445 to make sure it's consistent (you only have push_sandbox down there)

davisagli: Also now I'm thinking a better way to determine whether to make these calls is if the org is named `packaging`. That way we'll be sure to require CTC checks for any plan that runs against production.

We should add a Django setting based on an environment variable for whether to enable this feature. That way we can merge it even if we're not ready to enable it in production yet.

metaci/release/utils.py
```diff
@@ -83,10 +83,101 @@ def send_release_webhook(project_config, release):
...
+        f"Sending start webhook for {release} to {settings.METACI_RELEASE_WEBHOOK_URL}"
+    )
+    if role == "push_sandbox":
+        implementation_step_id = release.implementation_steps.get(
+            plan__role="push_sandbox"
+        ).implementation_step_id
+    if role == "push_production":
+        implementation_step_id = release.implementation_steps.get(
+            plan__role="push_production"
+        ).implementation_step_id
```
davisagli, line 117: Instead of looking specifically for these roles, you could query for implementation steps with a matching role:

    implementation_step_id = release.implementation_steps.get(plan__role=role)

metaci/release/utils.py
```diff
@@ -83,10 +83,101 @@ def send_release_webhook(project_config, release):
...
+        "implementation_step_id": f"{implementation_step_id}",
+    }
+    token = jwt.encode(
+        {
+            "iss": settings.METACI_RELEASE_WEBHOOK_ISSUER,
+            "exp": timegm(datetime.utcnow().utctimetuple()),
+        },
+        settings.METACI_RELEASE_WEBHOOK_AUTH_KEY,
+        algorithm="HS256",
+    )
```
davisagli, line 128: We should factor the code for generating the JWT out into a helper function instead of repeating it.

metaci/release/utils.py
```diff
@@ -83,10 +83,101 @@ def send_release_webhook(project_config, release):
...
+    response = requests.post(
+        f"http://0.0.0.0:8001/implementation_step_id/{implementation_step_id}/start/",
+        json=payload,
+        headers={"Authorization": f"Bearer {token}"},
+    )
+    result = response.json()
+    if result["success"]:
+        #######################################################################
+        # PARSE RESULT HERE FOR IMPLMENTATION ID's when GUS SANDBOX is ready! #
```
davisagli, line 137: We already have the ids; I don't think we have to do anything if the start was successful.

metaci/release/utils.py
```diff
@@ -83,10 +83,101 @@ def send_release_webhook(project_config, release):
...
+        f"Sending start webhook for {release} to {settings.METACI_RELEASE_WEBHOOK_URL}"
+    )
+    if role == "push_sandbox":
+        implementation_step_id = release.implementation_steps.get(
+            plan__role="push_sandbox"
+        ).implementation_step_id
+    if role == "push_production":
+        implementation_step_id = release.implementation_steps.get(
+            plan__role="push_production"
+        ).implementation_step_id
```
davisagli, line 158: Same suggestion re querying for a matching role

metaci/release/utils.py
```diff
@@ -83,10 +83,101 @@ def send_release_webhook(project_config, release):
...
+        json=payload,
+        headers={"Authorization": f"Bearer {token}"},
+    )
+    result = response.json()
+    if result["success"]:
+        #######################################################################
+        # PARSE RESULT HERE FOR IMPLMENTATION ID's when GUS SANDBOX is ready! #
+        #######################################################################
+        with transaction.atomic():
+            release.save()
```
davisagli, line 181: I don't think we need to do anything if the stop was successful.

metaci/release/utils.py
```diff
@@ -55,9 +55,9 @@ def update_release_from_github(release, repo_api=None):
     return release
 
 
-def send_release_webhook(project_config, release):
-    if release is None or not settings.METACI_RELEASE_WEBHOOK_URL:
-        return
+def send_release_webhook(project_config, release, config_item):
+    if release is None or not settings.METACI_RELEASE_WEBHOOK_URL or not settings.GUS_BUS_ENABLED or not config_item:
```
davisagli, line 59: The point of the new setting is to give us a way to keep the existing gus-bus integration enabled (for creating the change case) but not enable the start/stop calls yet. So we should not check the new setting here. If we need a way to toggle all of the webhooks at once, we can already do that using METACI_RELEASE_WEBHOOK_URL.

Also for now we still need to send the release webhook and create a change case even if we don't have a config item. We can update it to be stricter later once we've set up the config item path for all our orgs.

metaci/release/utils.py
```diff
@@ -67,6 +67,52 @@ def send_release_webhook(project_config, release):
...
+                "start_time": release.implementation_steps.get(
+                    plan__role="push_production"
+                ).start_time.isoformat(),
+                "end_time": release.implementation_steps.get(
+                    plan__role="push_production"
+                ).stop_time.isoformat(),
+                "implementation_steps": "push_production",
+                "configuration_item": config_item
+            },
+        ],
```
davisagli, line 115: We should:
1. Only add these steps if config_item has a value.
2. Instead of a hardcoded list of implementation steps, we should create a list from release.implementation_steps.all() and loop over it to add each of those. That way it's possible to customize the steps for a particular release if necessary.

metaci/release/utils.py
```diff
@@ -83,10 +129,98 @@ def send_release_webhook(project_config, release):
...
+            )
+            release.implementation_steps.filter(plan__role="release").update(
+                external_id=result["implementationSteps"][1]
+            )
+            release.implementation_steps.filter(plan__role="push_sandbox").update(
+                external_id=result["implementationSteps"][2]
+            )
+            release.implementation_steps.filter(plan__role="push_production").update(
+                external_id=result["implementationSteps"][3]
+            )
```
davisagli, line 148: This part needs to be inside a conditional for the new setting so that we don't try to run it in production yet. It can refer back to the list of implementation steps created above so that we don't need to re-query.

metaci/release/utils.py
```diff
@@ -83,10 +129,98 @@ def send_release_webhook(project_config, release):
...
+    response = requests.post(
+        f"http://0.0.0.0:8001/implementation_step_id/{implementation_step_id}/stop/",
+        json=payload,
+        headers={"Authorization": f"Bearer {token}"},
+    )
+    result = response.json()
+    if result["success"]:
+        #######################################################################
+        # PARSE RESULT HERE FOR IMPLMENTATION ID's when GUS SANDBOX is ready! #
+        #######################################################################
```
davisagli, line 221: Remove this comment which is no longer relevant

davisagli (changes requested):

metaci/release/utils.py
```diff
@@ -55,38 +78,131 @@ def update_release_from_github(release, repo_api=None):
...
         settings.METACI_RELEASE_WEBHOOK_URL,
         json=payload,
         headers={"Authorization": f"Bearer {token}"},
     )
     result = response.json()
     if result["success"]:
+        result["implementationSteps"] = ["1", "2", "3", "4"]
         with transaction.atomic():
+            if settings.METACI_START_STOP_WEBHOOK:
+                for i in range(0, len(implementation_steps) - 1):
```
jofsky, line 116: @davisagli There has to be a better way.

metaci/release/utils.py
```diff
@@ -55,27 +78,31 @@ def update_release_from_github(release, repo_api=None):
...
     tag = release.git_tag
+
+    steps = []
+    if config_item and settings.METACI_START_STOP_WEBHOOK:
+        implementation_steps = release.implementation_steps.all()
+        for implementation_step in implementation_steps:
+            steps = [
+                implementation_payload(implementation_step.plan.role, config_item)
+                for implementation_step in release.implementation_steps.all()
+            ]
```
davisagli, line 96:
```suggestion
        steps = [
            implementation_payload(implementation_step.plan.role, config_item)
            for implementation_step in implementation_steps
        ]
```

metaci/release/utils.py
```diff
@@ -13,6 +13,29 @@
...
+            "description": role,
+            "start_time": role,
+            "end_time": role,
+            "configuration_item": config_item,
+            "implementation_steps": role,
+        }
+    raise Exception("Please check your plan's role and org's configuration item.")
+
+
+def jwt_session():
```
davisagli, line 28: Maybe call this jwt_for_webhook() -- unlike the jwt_session function in CumulusCI that's used to get a JWT for authenticating to an org, this one does not get an access token back to start a session. And using a different name will help make it clearer that it's a different JWT.

metaci/release/utils.py
davisagli: How about:

    if "implementation_steps" in result:
        for step_model, step_result in zip(implementation_steps, result["implementationSteps"]):
            step_model.external_id = step_result
            step_model.save()

metaci/release/utils.py
```diff
@@ -84,9 +111,84 @@ def send_release_webhook(project_config, release):
...
+    )
+    implementation_step_id = release.implementation_steps.get(
+        plan__role=role
+    ).external_id
+    payload = {
+        "implementation_step_id": f"{implementation_step_id}",
+    }
+    token = jwt_session()
+    response = requests.post(
+        f"http://0.0.0.0:8001/implementation_step_id/{implementation_step_id}/start/",
```
davisagli, line 149: This should use settings.METACI_RELEASE_WEBHOOK_URL instead of a hardcoded local host/port

metaci/release/utils.py
```diff
@@ -143,9 +144,9 @@ def send_start_webhook(project_config, release, role, config_item):
     payload = {
         "implementation_step_id": f"{implementation_step_id}",
     }
-    token = jwt_session()
+    token = jwt_for_webhook()
     response = requests.post(
-        f"http://0.0.0.0:8001/implementation_step_id/{implementation_step_id}/start/",
+        f"http://{settings.METACI_RELEASE_WEBHOOK_URL}/implementation_step_id/{implementation_step_id}/start/",
```
davisagli, line 149: The setting should include https.

Also I meant for you to make the same change on line 184

metaci/release/utils.py
```diff
@@ -55,38 +78,117 @@ def update_release_from_github(release, repo_api=None):
...
-            "iss": settings.METACI_RELEASE_WEBHOOK_ISSUER,
-            "exp": timegm(datetime.utcnow().utctimetuple()),
-        },
-        settings.METACI_RELEASE_WEBHOOK_AUTH_KEY,
-        algorithm="HS256",
-    )
+    token = jwt_for_webhook()
     response = requests.post(
-        settings.METACI_RELEASE_WEBHOOK_URL,
+        f"{settings.METACI_RELEASE_WEBHOOK_URL}/release/",
```
davisagli, line 106: test needs to be adjusted to match

You do run the tests before committing, right?

jofsky: Usually I do but I forgot to late last night apologies.

davisagli: Sorry, that was a passive-aggressive way to ask about it. It's not like I always remember either.

jofsky: Lol it's ok I may have had a glass of wine or two last night, so it's totally warranted 🤣

metaci/release/models.py
```diff
@@ -79,6 +104,9 @@ def get_production_date():
     change_case_link = models.URLField(
         _("change case link"), max_length=1024, null=True, blank=True
     )
+    submitted_case = models.BooleanField(
+        _("submitted_case"), default=False, editable=False
+    )
```
davisagli, line 109: @jofsky I thought we decided we don't need a submitted_case field, since we can tell whether a case was submitted based on whether it was recorded in the change_case_link field

metaci/release/utils.py
```diff
@@ -55,38 +78,117 @@ def update_release_from_github(release, repo_api=None):
     return release
 
 
-def send_release_webhook(project_config, release):
+def send_release_webhook(project_config, release, config_item=None):
     if release is None or not settings.METACI_RELEASE_WEBHOOK_URL:
-        return
+        return  # should we better error handle this?
```
davisagli, line 83: This is not an error condition; we're intentionally keeping it possible to run an instance of MetaCI that isn't integrated with Change Traffic Control.

metaci/release/utils.py
```diff
@@ -55,38 +78,117 @@ def update_release_from_github(release, repo_api=None):
...
-        settings.METACI_RELEASE_WEBHOOK_AUTH_KEY,
-        algorithm="HS256",
-    )
+    token = jwt_for_webhook()
     response = requests.post(
-        settings.METACI_RELEASE_WEBHOOK_URL,
+        f"{settings.METACI_RELEASE_WEBHOOK_URL}/release/",
         json=payload,
         headers={"Authorization": f"Bearer {token}"},
     )
```
davisagli, line 109: Note to self: I need to be sure to adjust METACI_RELEASE_WEBHOOK_URL to remove /release/ when I deploy this.

metaci/release/utils.py
```diff
@@ -55,38 +78,117 @@ def update_release_from_github(release, repo_api=None):
...
+            implementation_payload(implementation_step.plan.role, config_item)
+            for implementation_step in implementation_steps
+        ]
+
     payload = {
         "case_template_id": release.change_case_template.case_template_id,
         "package_name": project_config.project__package__name,
         "version": project_config.get_version_for_tag(tag),
         "release_url": f"{release.repo.url}/releases/tag/{urllib.parse.quote(tag)}",
+        "steps": steps,
```
davisagli, line 102: Note to self: I need to be sure to deploy the updates to gus-bus at the same time that I deploy this.

davisagli (approved):

PR #2040, opened 2021-03-02
Import data supplied to perf test keywords

prescod:
Thiis keyword:

https://cumulusci.readthedocs.io/en/latest/robot.html?highlight=set%20elapsed%20time#set-test-elapsed-time

Adds some markup to output.xml.

The current PR parses out that data (if its available) and uses it as the duration instead of computing the duration the old way.


davisagli (approved):
LGTM

PR #2041, opened 2021-03-02
🛢 Add weekly and monthly plan schedules

prescod:
Title is self-explanatory.

Django tried to make two more migrations ('user', and 'site') but I thought it was safer not to include them in an unrelated PR.


davisagli (approved):
Looks fine, but I wonder if we should be aiming to provide better configurability of when individual scheduled planschedules run (i.e. set the hour for a daily build, set the day for a monthly build or DOW for a weekly build -- on the database model).

prescod:
I agree that that's another feature we should add.

PR #2042, opened 2021-03-03
One-off-dynos

prescod:
Changes:

* Allows Plans to ask for builds to be executed on one-off dynos instead of persistent worker dynos

Implementation:

* There is a "long-running" pseudo-queue in the UI and the DB which does not correspond to an RQ
* Jobs detected through that queue are dispatched through the AutoScaler to a new environment

When Pipeline API is available, the long-running queue could dispatch to that instead.


prescod:
If you get a chance to review if this strategy makes sense it would be helpful @davisagli

metaci/build/autoscaling.py
```diff
@@ -111,6 +114,20 @@ def scale(self):
                         )
                     )
 
+    def one_off_build(self, build_config):
+        config_args, no_lock = build_config[0:-1], build_config[-1]
```
davisagli, line 118:
```suggestion
        *config_args, no_lock = build_config
```

davisagli: (I was kinda surprised this works but it appears to)

metaci/build/autoscaling.py
```diff
@@ -111,6 +114,20 @@ def scale(self):
...
+        if no_lock:
+            config_args.extend(("--no-lock"))
+        proc = subprocess.Popen(
+            [
+                "python",
+                "./manage.py",
+                "metaci_rqworker",
```
davisagli, line 125: This should be the run_build command, right? Doesn't make much sense otherwise.

metaci/build/autoscaling.py
```diff
@@ -180,6 +198,25 @@ def scale_max(self, url, headers, worker_type, limit):
...
 
+    def one_off_build(self, build_config):
+        config_args, no_lock = build_config[0:-1], build_config[-1]
+        if no_lock:
+            config_args.extend(("--no-lock"))
+        command = " ".join(
+            [
+                "python",
+                "./manage.py",
+                "metaci_rqworker",
```
davisagli, line 209: Ditto re run_build

metaci/build/autoscaling.py
```diff
@@ -180,6 +198,25 @@ def scale_max(self, url, headers, worker_type, limit):
...
+                *config_args,
+            ]
+        )
+        url = f"{self.base_url}/dynos"
+        json = {"command": command, "time_to_live": "86400"}
+
+        resp = requests.post(url, json=json, headers=self.headers)
+        resp.raise_for_status()
```
davisagli, line 217: Can we do this in a way that makes the response from the Platform API visible to us? i.e. it's not terribly helpful to know that there was a 500 error without seeing the response body

metaci/plan/models.py
```diff
@@ -43,6 +43,7 @@
     ("medium", "medium priority"),
     ("high", "high priority"),
     ("robot", "robot tests"),
+    ("perftests", "performance tests"),
```
davisagli, line 46: I think this queue should be named based on how it operates (i.e. oneoff) rather than a particular use case. It'll be tied to performance tests by virtue of being the queue selected for performance test plans.

metaci/build/tasks.py
```diff
@@ -229,3 +239,31 @@ def delete_scratch_org(org_instance_id):
...
+
+
+class BuildSpec(T.NamedTuple):
+    repo_name: str
+    branch_name: str
+    commit: str
+    plan_name: str
+    username_or_email: str
+    timeout: int
+    no_lock: bool = False
```
davisagli, line 251: Not sure I see the benefit of creating a namedtuple for this?

metaci/build/tasks.py
```diff
@@ -229,3 +239,31 @@ def delete_scratch_org(org_instance_id):
...
+    build_config = BuildSpec(
+        repo=build.repo.name,
+        branch_name=build.branch_name,
+        commit=build.commit,
+        plan=build.plan.name,
+        username_or_email=build.org.name,
+        timeout=build.plan.build_timeout,
+    )
+
+    auto_scaler_app_name = list(settings.AUTOSCALERS.values())[1]
```
davisagli, line 264: Using AUTOSCALERS for this doesn't make much sense to me, since it's a list of apps that should be scaled based on queue size, which is _not_ what we're doing for the one-off dynos. I think it should be a new setting with the name of the app to use for one-off dynos.

metaci/build/tasks.py
```diff
@@ -93,14 +94,23 @@ def run_build(build_id, lock_id=None):
...
-    build.save()
-    autoscale()
+    queue_name = build.plan.queue
+    if queue_name == "perftest":
+
+        class Result(T.NamedTuple):
+            id: T.Any
+
+        job_id = run_one_off_build(build)
+        return Result(job_id)
```
davisagli, line 104: If there's an error here we should make sure that gets recorded on the build model and its status updated to error

metaci/build/tasks.py
```diff
@@ -93,14 +94,23 @@ def run_build(build_id, lock_id=None):
...
-    )
-    build.task_id_check = None
-    build.task_id_run = result.id
-    build.save()
-    autoscale()
+    queue_name = build.plan.queue
+    if queue_name == "perftest":
+
+        class Result(T.NamedTuple):
+            id: T.Any
```
davisagli, line 101: Do we need to return anything? And if so why a namedtuple?

metaci/build/tasks.py
davisagli: Would probably also be good to log what dyno it's running in when it's successful, so that we can find it if we need to kill it.

davisagli (commented):
Needs some work on details but I think the overall approach is sound.

metaci/build/autoscaling.py
prescod, line 125:
```suggestion
                "run_build",
```

metaci/build/autoscaling.py
prescod, line 209:
```suggestion
                "run_build",
```

metaci/build/tasks.py
prescod: We need to return an object with an "id" property because upstream code is expecting to log the "result.id" which I guess is the Django job id.

metaci/build/tasks.py
prescod: Do you want me to pull the code all of the way out of the Autoscalers despite it being "similar" to autoscaler code (in that it uses the same endpoints as the autoscaler)? Or do you just want to make it *possible* to config the One-of creator to be separable from the autoscaler?

prescod: Or do you want me to take another attempt at making a "job runner API interface" class which is separate from the autoscaler logic?

davisagli: @prescod Oh, I'm okay with having one_off_build be a method on the autoscaler class -- that's a convenient place to put it for swapping the implementation between local and Heroku. I just meant that I think it should get the name of the app to use for one-off dynos from a new setting, not from the AUTOSCALERS setting, because it seems pretty arbitrary that you're pulling out the 2nd entry from AUTOSCALERS. (Also I think the values are dicts, not an app name.)

metaci/build/management/commands/queue_build.py
```diff
@@ -0,0 +1,37 @@
+from django.core.management.base import BaseCommand
+
+from metaci.build.management.commands._utils import make_build
+
+
+class Command(BaseCommand):
```
Br4nd0R, line 6: @prescod - You mentioned maybe delete this?

Br4nd0R:
@prescod - This looks good to me, thanks for meeting to chat about it. Glad we caught the unneeded command! (I'll let David G. formally approve)

metaci/build/tasks.py
```diff
@@ -92,8 +96,39 @@ def run_build(build_id, lock_id=None):
...
 
-def start_build(build, lock_id=None):
-    queue = django_rq.get_queue(build.plan.queue)
+def dispatch_build(build, lock_id: str = None):
+    queue_name = build.plan.queue
+    if queue_name == "long-running":
+        return dispatch_one_off_build(build, lock_id)
+    else:
+        return dispatch_queued_build(build, lock_id)
+
```
davisagli, line 105: I'm confused by your approach here.

If you want long-running builds to use check_queued_build in the short queue to make sure that we have an org available (scratch org capacity or a persistent org lock), then how will the lock get released at the end of metaci.build.tasks.run_build? It won't have the lock id.

On the other hand if you want to check to make sure we have an org available when the one-off dyno starts up (like you have in here https://github.com/SFDO-Tooling/MetaCI/pull/2042/files#diff-894872607343abb7ad0807783377d82e2807ec5d8561b9a659fc5e0677280673R48) then I think we may as well call run_one_off_build directly from metaci.build.handlers.queue_build when the Build model is saved, and not go through check_queued_build.

metaci/cumulusci/logger.py
```diff
@@ -21,7 +22,9 @@ def __init__(self, model):
         self.buffer = ""
         self.last_save_time = timezone.now()
 
-    def flush(self, force=False):
+    # TODO: Force mode seems to be always used.
+    #       Is that intended?
```
davisagli, line 26: It seems to be.

metaci/cumulusci/logger.py
```diff
@@ -49,14 +52,20 @@ def init_logger(model):
...
     logger = logging.getLogger("cumulusci")
 
-    # Remove existing handlers
-    for handler in list(logger.handlers):
-        handler.stream.flush(force=True)
-        logger.removeHandler(handler)
+    if os.environ.get("LOG_TO_STDERR"):
+        handler = logging.StreamHandler()
+        print("Logging to STDERR")
+        model.log = "Logging to STDERR"
```
davisagli, line 58: I can see the use of this for debugging locally, but it should probably log to stderr _in addition_ to the db model, rather than instead of it.

metaci/plan/migrations/0013_create_default_plans.py
```diff
@@ -87,6 +87,15 @@ def create_default_plans(apps, schema_editor):
...
     )
 
+    Plan.objects.create(
+        name="Performance Test",
+        description=("Runs the test_performance_scratch flow aginst a scratch org"),
+        type="manual",
+        flows="test_performance_scratch",
+        org="dev",
+        public=False,
+    )
```
davisagli, line 97: Is the goal here to create a Performance Test plan for all new instances of MetaCI only? Or to add one to all existing instances of MetaCI when they upgrade?

Should this use the long-running queue?

davisagli (changes requested):

metaci/build/management/commands/queue_build.py
prescod: 04d6c070

config/settings/production.py
```diff
@@ -273,6 +273,7 @@
         }
     }
 
+LONG_RUNNING_BUILD_SCALER = env("LONG_RUNNING_BUILD_SCALER", default=HEROKU_APP_NAME)
```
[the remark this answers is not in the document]
prescod: 8f1140b6

config/settings/test.py
```diff
@@ -89,3 +89,5 @@
         "queues": ["default", "medium", "high"],
     }
 }
+
+LONG_RUNNING_BUILD_SCALER = "test-app"
```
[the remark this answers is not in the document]
prescod: 8f1140b6

integration-tests/run-one-off-build.py
```diff
@@ -0,0 +1,46 @@
...
+        UserFactory(username="fake_username")
+        build = make_build(
+            "CumulusCI-Test",
+            "main",
+            "main",
+            "myplan",
+            "fake_username",
+        )
+        c = Command()
+        c.handle(build_id=build.id)
```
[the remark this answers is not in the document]
prescod: 669b8a8 and f0a2256

metaci/build/management/commands/run_build_from_id.py
```diff
@@ -0,0 +1,13 @@
+from django.core.management.base import BaseCommand
+
+from metaci.build.management.commands._utils import do_build
+
+
+class Command(BaseCommand):
+    help = "Runs a build from the command line (on local computer)."
```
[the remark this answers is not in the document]
prescod: By the time it runs, the place it is running is "the local computer", I think. The point is that it doesn't add to a queue. It just runs.

metaci/build/tasks.py
```diff
@@ -134,14 +169,17 @@ def check_queued_build(build_id):
         # For scratch orgs, we don't need concurrency blocking logic,
         # but we need to check capacity
 
+        # TODO: isn't there a very large race condition relating to
+        # checking for scratch org reserve before we even queue the
+        # build on the worker queue?
```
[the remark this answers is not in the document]
prescod: When we redo this on top of Pipeline API, I think it would be better to catch the specific failure and re-queue instead of just having a race condition which probably generates unrecoverable failures.

metaci/plan/migrations/0013_create_default_plans.py
prescod: New instances and tests. And yes.

metaci/build/tasks.py
prescod: Okay, I had some misunderstandings about how locking works and in fact the old run_build task probably didn't handle locking properly I think. (hopefully fixed now)

Still confused about the difference between `metaci.build.management.commands.run_build.lock_org` and `metaci.cumulusci.org.Org.lock()` but that's not relevant.

metaci/build/autoscaling.py
```diff
@@ -180,6 +203,31 @@ def scale_max(self, url, headers, worker_type, limit):
...
+            + no_lock_arg
+        )
+        url = f"{self.base_url}/dynos"
+        json = {"command": command, "time_to_live": "86400"}
+
+        resp = requests.post(url, json=json, headers=self.headers)
+        if resp.status_code != 200:
+            msg = f"One-off dyno could not be started: {resp.status_code}: {resp.reason} : {resp.text}"
+            logger.error(msg)
+            raise HTTPError(resp.url, resp.status_code, msg, resp.headers, resp)
```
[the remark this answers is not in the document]
prescod: 890b6c7b

metaci/build/tasks.py
prescod: 5ec4ac70

metaci/plan/migrations/0013_create_default_plans.py
prescod: bc7fe79db

integration-tests/test_run_one_off_build.py
```diff
@@ -0,0 +1,49 @@
...
+            "main",
+            "main",
+            "myplan",
+            "fake_username",
+        )
+        c = Command()
+        c.handle(build_id=build.id, lock_id="abcde")
+        outerr = capsys.readouterr()
+        print(outerr.err)
+        assert "ci_test_concurrency" in outerr.err + outerr.out
```
davisagli, line 49: I'm still not clear on the goal of this test. If you want to make sure the run_build_from_id command ran successfully, looking for the flow name in the output doesn't seem like a good indicator of that.

I can see the rationale for having an integration test that actually fetches the commit from github, but I'm not thrilled about adding a 60-second sleep to every MetaCI test run.

metaci/build/tasks.py
```diff
@@ -229,3 +264,17 @@ def delete_scratch_org(org_instance_id):
...
+
+
+def launch_one_off_build_worker(build, lock_id: str):
+    """Immediately launch a one-off-build with env-appropriate autoscaler"""
+    app_name = settings.METACI_LONG_RUNNING_BUILD_APP
+    assert (
+        app_name
+    ), "Need to define METACI_LONG_RUNNING_BUILD_APP to run one-off-builds"
+    autoscaler_class = import_global(settings.METACI_WORKER_AUTOSCALER)
+    autoscaler = autoscaler_class(settings.AUTOSCALERS[app_name])
```
davisagli, line 276:
```suggestion
    autoscaler = autoscaler_class({"app_name": settings.METACI_LONG_RUNNING_BUILD_APP, "worker_type": "not_used", "queues": [], "max_workers": 0, "worker_reserve": 0})
```

That way there's no need for a new entry in the AUTOSCALERS setting.

But the more I look at this, the more it feels like a one-off build dispatcher is a different type of component from an autoscaler. They both need an app name to target and they both have different implementations between local and Heroku, but aside from an app name to target they don't use any of the same configuration.

integration-tests/test_run_one_off_build.py
prescod: Ultimately, before I deploy a feature I want to see it run on my computer. This is a one-command way of seeing the thing work with my own eyes. I added it to the Github action because I anticipated a potential concern that a test that isn't run often will tend to rot. The integration test does not run locally when you just run 'pytest'. I don't feel strongly about whether it should or shouldn't run on every push to github, but I see relatively minor harm in it.

I could make the condition it tests for (slightly) more robust but the build doesn't execute all of the way to the end due to key management issues which I don't have the time to debug and are fairly orthogonal to what I want to test.

metaci/build/tasks.py
prescod: They are both related to scaling the app up and they will tend to use the same API endpoints and protocols (e.g. Heroku API vs. Boto vs. whatever). 

Also, this will all get deleted when we move to Pipeline API, so perhaps we shouldn't overthink it. It's not a central abstraction.

integration-tests/test_run_one_off_build.py
prescod: I made the assertion slightly more robust.

Also, the test only takes 11 seconds to run, including the 8-9 seconds of startup time for MetaCI tests in general, so I don't think there is any significant cost to it.

davisagli: @prescod What sort of key management issues? But it sounds like it's a test for confirming that the build started, not that it ran to completion.

metaci/build/tasks.py
davisagli: @prescod I'd still like to make the change I proposed here, so that it's not necessary to add a new entry to AUTOSCALERS for the long-running app. It feels wrong to force someone who wants to use it to pick values for 4 settings that are not actually going to be used.

Btw, METACI_LONG_RUNNING_BUILD_APP still needs to be mentioned in docs/configuring.rst and/or the section of the readme about autoscaling, right?

davisagli:
@prescod The tests failed with: `Conflicting migrations detected; multiple leaf nodes in the migration graph: (0037_auto_20210304_0908, 0037_auto_20210315_2204 in plan).`

You need to merge the migrations from this branch with the ones that got added to main when I merged Josh's work. https://www.algotech.solutions/blog/python/django-migrations-and-how-to-manage-conflicts/

config/settings/local.py
```diff
@@ -119,7 +119,8 @@
     "localhost:8080",
     "0.0.0.0",
 ]
-
+METACI_LONG_RUNNING_BUILD_CLASS = "metaci.build.autoscaling.LocalOneOffBuilder"
+METACI_LONG_RUNNING_BUILD_CONFIG = {}
```
davisagli, line 123: This is the default for METACI_LONG_RUNNING_BUILD_CONFIG from base.py; no need to override

config/settings/production.py
```diff
@@ -257,9 +257,13 @@
 
 if HEROKU_TOKEN and HEROKU_APP_NAME:
     METACI_WORKER_AUTOSCALER = "metaci.build.autoscaling.HerokuAutoscaler"
+    METACI_LONG_RUNNING_BUILD_CLASS = "metaci.build.autoscaling.HerokuOneOffBuilder"
 
 # Autoscalers are defined per METACI_APP
 AUTOSCALERS = json.loads(env("AUTOSCALERS", default="{}"))
+METACI_LONG_RUNNING_BUILD_CONFIG = json.loads(
+    env("METACI_LONG_RUNNING_BUILD_CONFIG", default="{}")
+)
```
davisagli, line 266: This is the default for METACI_LONG_RUNNING_BUILD_CONFIG from base.py; no need to override

config/settings/local.py
```diff
@@ -119,7 +119,8 @@
     "localhost:8080",
     "0.0.0.0",
 ]
-
+METACI_LONG_RUNNING_BUILD_CLASS = "metaci.build.autoscaling.LocalOneOffBuilder"
```
davisagli, line 122: I would probably put this in base.py so that we can test the production settings locally. We've occasionally had reason to do that.

metaci/build/autoscaling.py
```diff
@@ -181,6 +201,36 @@ def scale_max(self, url, headers, worker_type, limit):
         return requests.patch(url, json={"quantity": target_workers}, headers=headers)
 
 
+class HerokuOneOffBuilder(OneOffBuilder):
+    """Build one-off builds Heroku worker dynos."""
```
davisagli, line 205:
```suggestion
    """Run a build in a Heroku one-off dyno."""
```

metaci/build/tasks.py
```diff
@@ -229,3 +266,17 @@ def delete_scratch_org(org_instance_id):
...
     else:
         return f"Failed to delete org instance #{ord.id}"
+
+
+def launch_one_off_build_worker(build, lock_id: str):
+    """Immediately launch a one-off-build with env-appropriate autoscaler"""
+    config = settings.METACI_LONG_RUNNING_BUILD_CONFIG
+    assert isinstance(
+        config, dict
+    ), "Need to define METACI_LONG_RUNNING_BUILD_CONFIG to run one-off-builds"
```
davisagli, line 276: METACI_LONG_RUNNING_BUILD_CONFIG is always a dict, the way it's set up in the settings. I think the validation that's relevant would be in HerokuOneOffBuilder to confirm there's an app_name in the config.

config/settings/local.py
prescod: How does putting this in base.py help us test the production settings locally?

davisagli: If you run `DJANGO_SETTINGS_MODULE="config.settings.production" ./manage.py runserver` locally, this line will not run because there's no HEROKU_TOKEN or HEROKU_APP_NAME: https://github.com/SFDO-Tooling/MetaCI/pull/2042/files#diff-b547313c08ca0949c654ff88d1c7f07a56d0f6c3076fa8152ab1243ca402fccaR260

So METACI_LONG_RUNNING_BUILD_CLASS will be unset and trying to run a one-off build will fail. If we move this line to base.py, then it would inherit the default and run it in a subprocess.

davisagli: I don't feel strongly about this one. We can change it later if we need it.

config/settings/local.py
prescod: 6795ee55

config/settings/production.py
prescod: 6795ee55

metaci/build/autoscaling.py
prescod: bad4d34a1

metaci/build/tasks.py
prescod: It's not always a dict because JSON parsing isn't strongly typed.

But I added the other suggestion you proposed.

bad4d34a1

davisagli: Okay, fair point.

metaci/build/tasks.py
davisagli: @prescod I'll approve this to merge and test on staging once the new settings are documented. Thanks for your patience with all the back and forth.

README.rst
```diff
@@ -124,14 +124,39 @@ To configure autoscaling:
...
 
+One-Off Builds
+~~~~~~~~~~~~~~
+
+In some environments, such as Heroku, it is helpful to run builds in
+environments which are spun up for just a single build. In Heroku, builds
+created in this way will not share their finite lifespan (24 hours) with
+previous builds. You might also want single-environment builds for
+security reasons, to ensure that no artifacts from one build can
+be snooped in another.
```
davisagli, line 143: One-off dynos do not help in that regard as long as they still have database credentials.

davisagli: The other benefit that's worth mentioning is that one-off dynos won't get restarted when the app is released.

README.rst
```diff
@@ -124,14 +124,39 @@ To configure autoscaling:
...
+You can specify the configuration for the class in JSON with an 
+environment variable called METACI_LONG_RUNNING_BUILD_CONFIG.
+
+For Heroku, this is a dictionary with a single key, like this:
+
+    METACI_LONG_RUNNING_BUILD_CONFIG = {"app_name": "my-app"}
+
+my-app would be replaced with the name of the Heroku App that should
+be used. It could be one of the apps you used for AUTOSCALERS but it
+need not be.
```
davisagli, line 158: I don't think things will work out very well if you try to use an app that is also being autoscaled. The one-off dynos consume a portion of the app's dyno limit, but the autoscaler is not counting one-off dynos so will think there is more capacity available than there is, and will try to scale to a value that will error. I think.

davisagli (approved):

PR #2049, opened 2021-03-04
🛢 Our dependency app migrations

prescod:
Generated when we run makemigrations.


davisagli (approved):
Thanks. I added a line to catch missing migrations in CI in the future.

PR #2097, opened 2021-03-22
[MetaCI] Release Form Change Creation Button

jofsky:


metaci/release/utils.py
```diff
@@ -96,8 +97,8 @@ def send_release_webhook(project_config, release, config_item=None):
 
     payload = {
         "case_template_id": release.change_case_template.case_template_id,
-        "package_name": project_config.project__package__name,
-        "version": project_config.get_version_for_tag(tag),
+        "package_name": release.repo.name,  # Need to figure out.
```
davisagli, line 100: This is probably okay. In the future we might want to add a field on Repository so that we can configure a project name that is different from the repo name.

metaci/templates/admin/release/release/change_form.html
```diff
@@ -0,0 +1,11 @@
+{% extends "admin/change_form.html" %}
+{% load i18n admin_urls static admin_modify %}
+{% block submit_buttons_bottom %}
+{% submit_row %}
+{% if METACI_RELEASE_WEBHOOK_URL and METACI_START_STOP_WEBHOOK == "True" %}
```
davisagli, line 5: I think this can just be `if METACI_RELEASE_WEBHOOK_URL` ... being able to create the change case can still be useful even if the option to do start/stop for implementation steps isn't enabled.

metaci/views.py
```diff
@@ -56,3 +55,10 @@ def get_context_data(self, **kwargs):
 def custom_403(request, exception):
     """Always redirect users to login via GitHub"""
     return HttpResponseForbidden(render(request, "account/login.html"))
+
+
+def gus_bus_enabled(request):
+    context = {
+        "METACI_START_STOP_WEBHOOK": os.environ.get("METACI_START_STOP_WEBHOOK", False)
+    }
+    return render(request, "admin/release/release/change_form.html", context=context)
```
davisagli, line 64: This function is no longer used and should be removed.

metaci/release/utils.py
```diff
@@ -90,14 +95,14 @@ def send_release_webhook(project_config, release, config_item=None):
...
+            implementation_payload(implementation_step.plan.role, config_item, release)
             for implementation_step in implementation_steps
         ]
 
     payload = {
         "case_template_id": release.change_case_template.case_template_id,
-        "package_name": project_config.project__package__name,
-        "version": project_config.get_version_for_tag(tag),
+        "package_name": release.repo.name,  # Need to figure out.
+        "version": release.version_number,  # Need to see if this is valid.
```
davisagli, line 105:
```suggestion
        "version": release.version_number,
```

metaci/release/utils.py
```diff
@@ -124,6 +129,32 @@ def send_release_webhook(project_config, release, config_item=None):
...
 
 
+def send_submit_webhook(release, config_item=None):
+    if (
+        release is None
+        or not settings.METACI_RELEASE_WEBHOOK_URL
+        or not settings.METACI_START_STOP_WEBHOOK
+        or not config_item
+    ):
+        return  # should we better error handle this?
```
davisagli, line 139:
```suggestion
        return
```

metaci/release/utils.py
```diff
@@ -124,6 +129,32 @@ def send_release_webhook(project_config, release, config_item=None):
...
+    logger.info(
+        f"Sending release webhook for {release} to {settings.METACI_RELEASE_WEBHOOK_URL}"
```
davisagli, line 141: This function is not sending a release webhook.

davisagli (changes requested):

metaci/release/admin.py
```diff
@@ -68,6 +70,26 @@ def get_form(self, request, obj=None, **kwargs):
     empty_value_display = "-empty-"
     inlines = [ImplementationStepInline]
 
+    def response_change(self, request, obj):
+        if (
+            "_create-change-case" in request.POST
+        ):  # should we do a try catch for better error handling?
```
davisagli, line 76:
```suggestion
        ):
```

metaci/release/admin.py
```diff
@@ -68,6 +70,26 @@ def get_form(self, request, obj=None, **kwargs):
...
+            send_release_webhook(
+                obj, obj.repo.orgs.get(name="packaging").configuration_item
+            )
+            self.message_user(request, "The change case has been successfully created.")
+        if (
+            "_submit-change-case" in request.POST
+        ):  # should we do a try catch for better error handling?
```
davisagli, line 83:
```suggestion
        ):
```

metaci/release/context_processors.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+
+def get_release_values(request):
+    return {
+        "METACI_START_STOP_WEBHOOK": os.environ.get("METACI_START_STOP_WEBHOOK", False),
```
davisagli, line 6:
```suggestion
```

metaci/release/utils.py
```diff
@@ -13,12 +13,17 @@
...
 
-def implementation_payload(role, config_item):
-    if role and config_item:
+def implementation_payload(role, config_item, release):
+    if role and config_item and release:
         return {
             "description": role,
-            "start_time": role,
-            "end_time": role,
+            "owner": "00XXXXXXXXXX",  # need to tie GUS user id to MetaCI user.
```
davisagli, line 20: I think we need to sort this out before we can merge this. Otherwise trying to create the case will fail, no?

metaci/release/utils.py
```diff
@@ -13,12 +13,17 @@
...
+            "start_time": release.implementation_steps.get(
+                plan__role=f"{role}"
```
davisagli, line 22: There's no reason to use an f-string if it's already a string

```suggestion
                plan__role=role
```

metaci/release/utils.py
```diff
@@ -13,12 +13,17 @@
...
+            ).start_time.isoformat(),
+            "end_time": release.implementation_steps.get(
+                plan__role=f"{role}"
```
davisagli, line 25:
```suggestion
                plan__role=role
```

metaci/release/utils.py
```diff
@@ -124,6 +129,32 @@ def send_release_webhook(project_config, release, config_item=None):
...
+        f"{settings.METACI_RELEASE_WEBHOOK_URL}/case/{release.change_case_link}/submit",
+        json=payload,
+        headers={"Authorization": f"Bearer {token}"},
+    )
+    result = response.json()
+    if result["success"]:
+        return
+    else:
+        raise Exception("\n".join(err["message"] for err in result["errors"]))
+
```
davisagli, line 156: send_submit_webhook doesn't have any test coverage

davisagli (dismissed):

metaci/release/utils.py
jofsky: That's because gus sandbox configuration doesn't support at the moment so it's blocked. Was going to write test coverage once we know the proper response payload.

metaci/release/utils.py
jofsky: @davisagli what do you propose passing it through a variable?

davisagli: @jofsky Do you think we can just use our GUS API user for all of them?

metaci/release/utils.py
davisagli: @jofsky ok, fair enough.

davisagli: @jofsky I don't think the actual response from GUS matters that much here. We just want to know whether it succeeded or failed. And errors should be reported by gus-bus in the same way as for other endpoints. So I think we could go ahead and write a test and probably not need to change it much later.

jofsky: Will do now.

metaci/release/utils.py
jofsky: I was hoping to map it to each ERE. Apply PLOP and only give those users in metaCI mappings if possible.

davisagli: Sure, that would be ideal but is more work so I'm wondering if we can use the API user for now, and then improve it later, rather than adding to the scope of this PR.

metaci/release/tests/test_utils.py
```diff
@@ -57,4 +57,52 @@ def test_send_release_webhook__error(mocked_responses, mocker, transactional_db)
...
+        "POST",
+        "https://webhook/case/None/submit",
+        json={"success": True, "id": "2"},
+    )
+
+    project_config = Mock(project__package__name="Test Package")
+    project_config.get_version_for_tag.return_value = "1.0"
+    release = ReleaseFactory()
+    send_submit_webhook(release, "INFRA.instance1")
+    assert not release.change_case_link
```
davisagli, line 81: This doesn't seem like the right behavior. If send_submit_webhook is called with a release that doesn't have a change_case_link yet, I would expect that to throw an error rather than calling gus-bus with "None" as the case id.

metaci/release/tests/test_utils.py
```diff
@@ -57,4 +57,52 @@ def test_send_release_webhook__error(mocked_responses, mocker, transactional_db)
...
+    )
+    mocked_responses.add(
+        "POST",
+        "https://webhook/case/None/submit",
+        json={
+            "success": False,
+            "errors": [
+                {"message": "ImplementationStep matching query does not exist."}
+            ],
+        },
```
davisagli, line 104: Based on what I see in gus-bus (https://github.com/SFDO-Tooling/gus-bus/blob/main/gus_bus/main.py#L76-L79), the expected response here would be:

```suggestion
        json={
            "success": False,
            "errors": ["Error submitting change case."],
        },
```

env.example
```diff
@@ -49,3 +49,10 @@ DJANGO_SERVER_EMAIL=
 # Sentry
 # Only needed if you want to send errors to Sentry
 DJANGO_SENTRY_DSN=
+
+
+# GUS BUS
+export METACI_RELEASE_WEBHOOK_URL=http://0.0.0.0:8001
+export METACI_RELEASE_WEBHOOK_AUTH_KEY=123645
+export METACI_START_STOP_WEBHOOK=True
+export GUS_BUS_OWNER_ID="005B00000000000000" # must edit yourself
```
davisagli, line 58: Let's keep these commented out in env.example. If someone else is trying to figure out how to run MetaCI locally I don't want them to get hung up because they aren't using the GUS integration.

metaci/build/models.py
```diff
@@ -428,9 +428,7 @@ def run(self):
 
         if self.plan.role == "release":
             try:
-                send_release_webhook(
-                    project_config, self.release, self.org.configuration_item
-                )
+                send_release_webhook(self.release, self.org.configuration_item)
```
davisagli, line 431: Currently we use this to create change cases automatically after running an Upload Release plan.

But since cases need to be approved before we can start their implementation steps, we are switching to creating them ahead of time using the Create Change Case button.

I think that means we need to remove lines 429-438 here so that we don't end up creating a duplicate case after the plan runs. Does that make sense?

metaci/release/tests/test_utils.py
```diff
@@ -13,22 +13,25 @@ def test_send_release_webhook(mocked_responses, mocker, transactional_db):
...
     )
 
     project_config = Mock(project__package__name="Test Package")
     project_config.get_version_for_tag.return_value = "1.0"
     release = ReleaseFactory()
 
-    send_release_webhook(project_config, release, "INFRA.instance1")
+    send_release_webhook(release, "INFRA.instance1")
 
     assert release.change_case_link == "2"
```
davisagli, line 30: Do we need an assertion to make sure that the ids from implementationSteps were stored on the correct ImplementationStep models? They need to be there so that the start/stop automation can find the right id.

metaci/release/tests/test_utils.py
```diff
@@ -57,4 +60,50 @@ def test_send_release_webhook__error(mocked_responses, mocker, transactional_db)
...
+        "POST",
+        "https://webhook/case/2/submit",
+        json={"success": True, "id": "2"},
+    )
+
+    project_config = Mock(project__package__name="Test Package")
+    project_config.get_version_for_tag.return_value = "1.0"
+    release = ReleaseFactory()
+    release.change_case_link = "2"
+    send_submit_webhook(release, "INFRA.instance1")
```
davisagli, line 84: This should have an assertion to confirm that the mocked response was called.

metaci/release/utils.py
```diff
@@ -13,12 +13,17 @@
...
 
-def implementation_payload(role, config_item):
-    if role and config_item:
+def implementation_payload(role, config_item, release):
+    if role and config_item and release:
         return {
             "description": role,
-            "start_time": role,
-            "end_time": role,
+            "owner": f"{settings.GUS_BUS_OWNER_ID}",
```
davisagli, line 20: There's no reason to use an f-string if all it does is include a variable that is already a string:

```suggestion
            "owner": settings.GUS_BUS_OWNER_ID,
```

metaci/release/utils.py
```diff
@@ -78,26 +83,28 @@ def update_release_from_github(release, repo_api=None):
     return release
 
 
-def send_release_webhook(project_config, release, config_item=None):
-    if release is None or not settings.METACI_RELEASE_WEBHOOK_URL:
+def send_release_webhook(release, config_item=None):
+    if (
+        release is None
+        or not settings.METACI_RELEASE_WEBHOOK_URL
+        or not settings.GUS_BUS_OWNER_ID
```
davisagli, line 90: GUS_BUS_OWNER_ID is required in order for us to send implementation steps, but it is not required for us to create a change case with no implementation steps. I don't think we want it included in the check here.

metaci/release/utils.py
```diff
@@ -124,11 +131,48 @@ def send_release_webhook(project_config, release, config_item=None):
...
+        else:
+            raise Exception("\n".join(err["message"] for err in result["errors"]))
+
+
+def send_start_webhook(release, role, config_item):
     if (
         release is None
         or not settings.METACI_RELEASE_WEBHOOK_URL
         or not settings.METACI_START_STOP_WEBHOOK
+        or not settings.GUS_BUS_OWNER_ID
```
davisagli, line 175: Don't need this check; the owner was set when the implementation step was created.

metaci/release/utils.py
```diff
@@ -159,11 +203,12 @@ def send_start_webhook(project_config, release, role, config_item):
         raise Exception("\n".join(err["message"] for err in result["errors"]))
 
 
-def send_stop_webhook(project_config, release, role, config_item):
+def send_stop_webhook(release, role, config_item):
     if (
         release is None
         or not settings.METACI_RELEASE_WEBHOOK_URL
         or not settings.METACI_START_STOP_WEBHOOK
+        or not settings.GUS_BUS_OWNER_ID
```
davisagli, line 211: Don't need this check; the owner was set when the implementation step was created.

env.example
```diff
@@ -49,3 +49,11 @@ DJANGO_SERVER_EMAIL=
...
 # Only needed if you want to send errors to Sentry
 DJANGO_SENTRY_DSN=
+
+
+# GUS BUS
+export METACI_RELEASE_WEBHOOK_URL=http://0.0.0.0:8001
+export METACI_RELEASE_WEBHOOK_AUTH_KEY=123645
+export METACI_START_STOP_WEBHOOK=True
+# Only needed when creating implementation steps find from GUS user id
+# export GUS_BUS_OWNER_ID="005B00000000000000"
```
davisagli, line 59: @jofsky I meant that we should comment out this entire section starting on line 55. Sending release and start/stop webhooks should be off by default.

metaci/release/tests/test_utils.py
```diff
@@ -18,15 +20,36 @@ def test_send_release_webhook(mocked_responses, mocker, transactional_db):
     mocked_responses.add(
         "POST",
         "https://webhook/release/",
-        json={"success": True, "implementationSteps": ["a2dAG0000000"], "id": "2"},
+        json={"success": True, "implementationSteps": ["2", "1", "3"], "id": "2"},
```
davisagli, line 23: This test is for a scenario with 2 implementation steps so we should only be expecting 2 ids to be returned, right?

jofsky: I wanted to check that the proper values were assigned to the implementation steps with the roles given.

davisagli: @jofsky I see 2 implementation steps set up on the release and 3 ids in the response. Why are they different?

davisagli (approved):
Thanks for all the iteration on this!

PR #2111, opened 2021-03-26
Improve how secrets are stored

davisagli:
# Important -- config updates required

We've made improvements to how a number of secrets used by MetaCI are stored. Some updates to config vars are needed before updating MetaCI to a commit that includes these changes:

1. The client id and client secret used for GitHub Oauth are now loaded from environment variables instead of stored in the SocialApp table in the database. Be sure to set GITHUB_CLIENT_ID and GITHUB_CLIENT_SECRET.
2. Org and service configuration is now encrypted before storing in the database. Run `python -c 'from cryptography.fernet import Fernet; print(Fernet.generate_key().decode("latin1"))'` to generate an encryption key, and add it to the DB_ENCRYPTION_KEYS environment variable before updating MetaCI with these changes. DB_ENCRYPTION_KEYS can be a newline-separated list of keys; the keys can be rotated by adding a new one at the beginning, but old keys need to remain in order to decrypt data that was stored using them.

---

Details:
- Migrated the JSON configuration field on Org, ScratchOrgInstance, and Service models to an EncryptedJSONField (based on django-fernet-fields). There's a data migration which is a bit complicated because it needs to add a new field, copy the data, then delete the old field and rename the new one.
- Added a customized version of the github oauth2 provider which ensures that the oauth client credentials can be loaded from settings. Based on prior work in Metecho. There's a migration to remove the existing client id and secret from the database.
- Added SOCIALACCOUNT_STORE_TOKENS = False setting to make django-allauth not store oauth access tokens in the database, since we aren't using them. There's a migration to delete existing tokens.

Incidental changes:
- Removed the CONNECTED_APP_CLIENT_ID, CONNECTED_APP_CLIENT_SECRET, and CONNECTED_APP_CALLBACK_URL settings, which were no longer actually used.
- Data in EncryptedJSONField is implicitly serialized to JSON, so there's no longer a need to explicitly call json.dumps/loads.
- Removed the ability to filter Orgs and Services by text contained in their json configuration.


app.json
```diff
@@ -11,16 +11,8 @@
...
-      "description": "The callback url configured for the Connected App.  You should be able to get these values from the command `cumulusci2 org connected_app` in your locally configured instance of CumulusCI",
-      "value": ""
-    },
-    "CONNECTED_APP_CLIENT_ID": {
-      "description": "The Client ID of the Salesforce Connected App.  You should be able to get these values from the command `cumulusci2 org connected_app` in your locally configured instance of CumulusCI",
-      "value": ""
-    },
-    "CONNECTED_APP_CLIENT_SECRET": {
-      "description": "The Client Secret of the Salesforce Connected App. You should be able to get these values from the command `cumulusci2 org connected_app` in your locally configured instance of CumulusCI",
+    "DB_ENCRYPTION_KEYS": {
```
jofsky, line 14: Why plural?

docs/configuring.rst
```diff
@@ -15,6 +15,13 @@ DJANGO_SECRET_KEY:
     This represents the secret key used to sign session cookies for the Django web application.
     Set it to an arbitrary string that is not shared with another Django site.
 
+DB_ENCRYPTION_KEYS:
```
jofsky, line 18: Ditto here.

env.example
```diff
@@ -3,6 +3,10 @@
 # Django secret key -- set this to an arbitrary string
 DJANGO_SECRET_KEY=
 
+# Encryption key for secrets stored in the database.
+# Generate a key using cryptography.fernet.Fernet.generate_key()
+DB_ENCRYPTION_KEYS=
```
jofsky, line 8: Ditto here.

metaci/fields.py
```diff
@@ -0,0 +1,125 @@
...
+        super(EncryptedField, self).__init__(*args, **kwargs)
+
+    @cached_property
+    def keys(self):
+        assert len(settings.DB_ENCRYPTION_KEYS) > 0
+        return settings.DB_ENCRYPTION_KEYS
+
+    @cached_property
+    def fernet(self):
+        return MultiFernet([Fernet(k) for k in self.keys])
```
jofsky, line 65: Is this why we have plural for keys?

app.json
jofsky: After reviewing metaci/tests/test_fields.py I think I understand why but want to use this to open up a brief discussion on this.

jofsky (approved):

app.json
davisagli: Does this sentence in the pull request description help clarify? "DB_ENCRYPTION_KEYS can be a newline-separated list of keys; the keys can be rotated by adding a new one at the beginning, but old keys need to remain in order to decrypt data that was stored using them."

PR #2112, opened 2021-03-26
Better logging

prescod:
Log is a string, not an array. It didn't crash but it didn't look right either.


davisagli (approved):
I should have caught this the first time!

And we should probably check if anyone has figured out a good way to add type hints to Django models.

PR #2147, opened 2021-04-05
Added migration to change testresults.robot_tags from CharField to TextField

boakley:
This, to remove an 80-character limit. I modifed some test data so that the test would have more tags than would fit in an 80 character field.


davisagli (approved):
I was going to leave this one for @prescod to review, but it looks fine to me and there's another PR with a migration from Josh and it seems fair to merge the one that was submitted earlier first so that @boakley doesn't have to merge the migrations.

PR #2148, opened 2021-04-06
🛢 [MetaCI] Handling case template id to be non nullable and migration…

jofsky:
- Require Release records to have a lookup to a Change Case Template. There's a migration that creates a Change Case Template named "Legacy Release" for existing releases that don't have one.
- Release records are no longer created automatically based on github webhooks for a release tag.


metaci/release/migrations/0014_auto_20210406_0241.py
```diff
@@ -0,0 +1,36 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name="release",
+            name="change_case_template",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.PROTECT,
+                to="release.changecasetemplate",
+            ),
+        ),
+    ]
```
davisagli, line 32: lines 23-32 are completely unused here, because you are replacing the `operations` list on line 34

metaci/release/migrations/0015_auto_20210407_0218.py
```diff
@@ -0,0 +1,22 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name="release",
+            name="change_case_template",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.PROTECT,
+                to="release.changecasetemplate",
+            ),
+        ),
```
davisagli, line 21: This doesn't look right -- it doesn't set null=False and blank=False, so the migrations end up with a final state that doesn't match the model in models.py.

Btw, I merged a PR from Bryan that added a different migration, so you will probably need to add a new merge migration so that the migration system can order the migrations correctly. (see "migration conflicts" in https://www.algotech.solutions/blog/python/django-migrations-and-how-to-manage-conflicts/)

metaci/repository/views.py
```diff
@@ -264,7 +264,11 @@ def get_or_create_release(event, tag, repo):
     release, _ = Release.objects.get_or_create(
         repo=repo,
         git_tag=tag,
-        defaults={"created_from_commit": event["head_commit"]["id"], "status": "draft"},
+        defaults={
+            "created_from_commit": event["head_commit"]["id"],
+            "status": "draft",
+            "change_case_template_id": 1,
+        },
```
davisagli, line 271: We aren't actually using this mechanism for creating a release based on a webhook from github, and I think it probably makes more sense to get rid of it rather than create a release with a default change case template which may or may not make sense.

davisagli (changes requested):

metaci/repository/views.py
jofsky: Removing this `"change_case_template_id": 1` causes metaci/repository/tests/test_views.py:385 to error out.

davisagli: I'm saying I think we should change this function so that it only finds existing releases but never creates a new one. We don't really want Release records to get created because of a github webhook, now that we are always creating them before a release tag gets created.

jofsky: I understand what you're saying but not sure what this would look like @davisagli

davisagli (approved):

PR #2155, opened 2021-04-07
Feature/automated push upgrades

jofsky:
This pull request enables push upgrade functionality in MetaCI.


metaci/build/models.py
```diff
@@ -713,6 +713,19 @@ def _get_flow_options(self) -> dict:
...
+        if (
+            self.build.plan.role == "push_sandbox" and self.build.release
+        ):  # override lives in MetaCI
+            options["push_sandbox"] = {
+                "version": f"{self.build.release.version_number}",
+            }
+        if (
+            self.build.plan.role == "push_production" and self.build.release
+        ):  # override lives in MetaCI
+            options["push_all"] = {
```
jofsky, line 725: @davisagli is it ok that we're using push_all? Theres no push_production task.

metaci/build/models.py
```diff
@@ -713,6 +713,19 @@ def _get_flow_options(self) -> dict:
                 "sandbox_date": self.build.release.sandbox_push_date,
                 "production_date": self.build.release.production_push_date,
             }
+        if (
+            self.build.plan.role == "push_sandbox" and self.build.release
+        ):  # override lives in MetaCI
+            options["push_sandbox"] = {
+                "version": f"{self.build.release.version_number}",
```
jofsky, line 720: This field isn't currently required, should I just parse the git tag? Or should I make a migration to make this required? @davisagli

davisagli (changes requested):
I think doing it this way is fine, but could you please add tests to confirm that the options are set correctly based on the plan role?

metaci/build/models.py
davisagli: Yeah, I think push_all is what we want, because on production Tuesday we want to include any new sandboxes that were created with the old version since sandbox Wednesday.

metaci/build/models.py
davisagli: We should probably make it required.

I think we could also potentially set defaults for both git tag and version number based on the most recent beta tag, but that should probably be a new work item and separate PR

davisagli (approved):

PR #2185, opened 2021-04-19
clarify that MetaCI doesn't support named services yet

davisagli:
- raise an error if cci tries to get a named service
- raise an error if cci tries to store a service


Br4nd0R (approved):

codexmarcel (approved):
Good call.

PR #2203, opened 2021-04-26
Feature/dynamic versioning

jofsky:


metaci/release/utils.py
```diff
@@ -36,6 +36,19 @@ def jwt_for_webhook():
...
+    if (
+        change_case_link
+        and len(
+            change_case_link.split(f"{settings.METACI_CHANGE_CASE_URL_TEMPLATE[:-9]}")
+        )
+        > 1
+    ):  # 500 to parse beginning sequence of change case; indexing [:-9] to remove "{case_id}" from url link
+        return f"500{change_case_link.split(settings.METACI_CHANGE_CASE_URL_TEMPLATE[:-9])[1]}"
+    else:
+        raise Exception("Please provide a valid change case link.")
```
davisagli, line 49: This looks pretty complicated and too closely tied to expecting METACI_CHANGE_CASE_URL_TEMPLATE to be in a particular format. I think we should:
- only store a case id in the change_case_link field, instead of a full URL like now
- construct the full URL from the id using METACI_CHANGE_CASE_URL_TEMPLATE when we need to

metaci/release/utils.py
```diff
@@ -96,8 +109,10 @@ def send_release_webhook(project_config, release, config_item=None):
 
     payload = {
         "case_template_id": release.change_case_template.case_template_id,
-        "package_name": project_config.project__package__name,
-        "version": project_config.get_version_for_tag(tag),
+        "package_name": release.repo.name,  # Need to figure out.
```
davisagli, line 112: I would stick with project_config.project__package__name here; package name is not necessarily the same as repo name.

metaci/release/utils.py
```diff
@@ -96,8 +109,10 @@ def send_release_webhook(project_config, release, config_item=None):
...
+        "version": release.repo.latest_release.git_tag.strip(
+            f"{release.repo.release_tag_regex}"
+        ),  # Need to see if this is valid.
```
davisagli, line 115: Just use release.git_tag, not release.repo.latest_release.git_tag. No one should be running the plan for an old release, but if they do we don't want to tell GUS that it's the latest release number.

And also, as written this will break if release.git_tag is null (release engineer forgot to fill in the field when creating the release). We need to either make the field required or make this code more robust so that it can raise an error that says what happened instead of just `AttributeError: NoneType object has no attribute strip`.

metaci/release/utils.py
```diff
@@ -107,6 +122,15 @@ def send_release_webhook(project_config, release, config_item=None):
...
         headers={"Authorization": f"Bearer {token}"},
     )
+    release.version_name = release.repo.latest_release.git_tag.strip(
+        f"{release.repo.release_tag_regex}"
+    )
+    release.version_number = release.repo.latest_release.git_tag.strip(
+        f"{release.repo.release_tag_regex}"
+    )
+    release.save()
+    breakpoint()
```
davisagli, line 132: Same comment re using release.git_tag instead of release.repo.latest_release.git_tag

metaci/release/utils.py
```diff
@@ -124,6 +148,32 @@ def send_release_webhook(project_config, release, config_item=None):
...
+def send_submit_webhook(release, config_item=None):
+    if (
+        release is None
+        or not settings.METACI_RELEASE_WEBHOOK_URL
+        or not settings.METACI_START_STOP_WEBHOOK
+        or not config_item
+    ):
+        return  # should we better error handle this?
+    logger.info(
+        f"Sending release webhook for {release} to {settings.METACI_RELEASE_WEBHOOK_URL}"
```
davisagli, line 160: wrong message for what this code is doing

metaci/views.py
```diff
@@ -56,3 +55,10 @@ def get_context_data(self, **kwargs):
 def custom_403(request, exception):
     """Always redirect users to login via GitHub"""
     return HttpResponseForbidden(render(request, "account/login.html"))
+
+
+def gus_bus_enabled(request):
```
davisagli, line 60: This function is not used anywhere

metaci/release/models.py
```diff
@@ -122,7 +122,16 @@ def update_from_github(self):
...
 
     def save(self, *args, **kw):
+        if not self.version_number:
+            self.version_number = self.repo.latest_release.git_tag.strip(
+                f"{self.repo.release_tag_regex}"
+            )
+        if not self.version_name:
+            self.version_name = self.repo.latest_release.git_tag.strip(
+                f"{self.repo.release_tag_regex}"
+            )
```
davisagli, line 132: What are you trying to do here? It looks like if a Release is saved that does not yet have a version name or number, they will get filled in by copying from the latest existing release. That doesn't sound right (we don't want to create a new release with the same number as one that already exists).

PR #2217, opened 2021-04-29
Week 17 dependency updates

Br4nd0R:


requirements/dev.txt
```diff
@@ -18,11 +18,11 @@ asgiref==3.3.4
     #   django
 attrs==20.3.0
     # via pytest
-babel==2.9.0
+babel==2.9.1
     # via sphinx
 backcall==0.2.0
     # via ipython
-black==20.8b1
+black==21.4b2
```
davisagli, line 25: When we update black we need to run it on existing files, and make sure that CI also runs the same version of black.

davisagli (changes requested):
Still need to update cumulusci, right?

PR #2230, opened 2021-05-05
Adjust cache configuration to work with Redis 6

davisagli:
Tell django-redis not to worry about a self-signed certificate when connecting with TLS, since that's what Heroku Redis 6 gives us

I also took the opportunity to consolidate the cache configuration into base.py

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -11,6 +11,7 @@ Production Configurations
 
 """
 import json
+import ssl
 
 from .base import *  # noqa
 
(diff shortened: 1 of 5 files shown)


config/settings/production.py
```diff
@@ -287,3 +268,9 @@
 
 # django-defender configuration
 DEFENDER_REDIS_NAME = "default"
+
+if REDIS_LOCATION.startswith("rediss://"):
```
Br4nd0R, line 272: Should this just have one 's'?

```suggestion
if REDIS_LOCATION.startswith("redis://"):
```

Br4nd0R (approved):

config/settings/production.py
davisagli: No, it's right. From https://devcenter.heroku.com/articles/heroku-redis:

> As of Redis version 6, Hobby plans support both TLS and unencrypted connections, while production plans require TLS connections. For hobby plans you can find the unencrypted (plaintext) and TLS URIs in your app’s config vars REDIS_URL and REDIS_TLS_URL respectively. For Premium, Private, and Shield plans on version 6 your REDIS_URL config var will use the rediss: scheme, and you must enable TLS in your Redis client’s configuration in order to connect to a Redis 6 database.

PR #2245, opened 2021-05-11
🛢 [MetaCI] Dynamic release form versioning

jofsky:


metaci/release/admin.py
```diff
@@ -10,6 +11,29 @@ class ImplementationStepInline(admin.TabularInline):
...
+                    form.base_fields["version_number"].initial = next_version
+                    form.base_fields[
+                        "git_tag"
+                    ].initial = f"{repo.release_tag_regex}{next_version}"
+            else:
+                form.base_fields["version_name"].initial = "1.0"
+                form.base_fields["version_number"].intial = "1.0"
+                form.base_fields[
+                    "git_tag"
+                ].initial = f"{repo.release_tag_regex}{next_version}"
```
davisagli, line 33: @jofsky I think we want something like this to get the version from the most recent beta release:

    # builds default to sorting with the most recent first
    latest_beta_tag = repo.builds.filter(plan__role="beta_test").first().branch.name
    # latest_beta_tag is something like "tag: uat/3.198-Beta_4"
    version = re.search('([\d\.]+)-Beta_\d+$', tag).group(1)

metaci/release/migrations/0016_auto_20210511_2018.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.1.7 on 2021-05-11 20:18
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    def populate_version_number(apps, schema_editor):
```
davisagli, line 7: It would be less confusing to put this as a standalone function outside the Migration class, since it doesn't take a `self` param and isn't called as an instance method or class method.

metaci/release/migrations/0016_auto_20210511_2018.py
```diff
@@ -0,0 +1,28 @@
...
+    def unpopulate_version_number(apps, schema_editor):
+        """Creating a NO-OP function for rollback migration capability."""
+        pass
+
+    dependencies = [
+        ("release", "0015_auto_20210407_0218"),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_version_number, unpopulate_version_number),
```
davisagli, line 20: You can do this so that you don't need to define an empty unpopulate_version_number:

```suggestion
        migrations.RunPython(populate_version_number, migrations.RunPython.noop),
```

metaci/release/models.py
```diff
@@ -61,7 +61,7 @@ class Release(StatusModel):
         _("version name"), max_length=255, null=True, blank=True
     )
     version_number = models.CharField(
-        _("version number"), max_length=255, null=True, blank=True
+        _("version number"), max_length=255, null=True, blank=False
```
davisagli, line 64: We should change both `null` and `blank` to False. A field with null=True and blank=False can still be set to null in code, which we don't want.

davisagli (changes requested):
Sorry for the slow feedback on this. I added a comment suggesting how to get the version number from the most recent beta.

metaci/release/admin.py
davisagli: @jofsky Now I'm having second thoughts about using betas, because it's possible that there might not yet be a new beta since the last final release at the point when the new release is planned. So let's go with your existing approach here.

davisagli, line 33: I would reorganize this slightly:

```suggestion
            next_version = None
            if repo.latest_release:
                if len(repo.latest_release.version_number.split(".")) > 1:
                    next_version = f"{repo.latest_release.version_number.split('.')[0]}.{int(repo.latest_release.version_number.split('.')[1]) + 1}"
            else:
                next_version = "1.0"
            if next_version is not None:
                form.base_fields["version_name"].initial = next_version
                form.base_fields["version_number"].initial = next_version
                form.base_fields[
                    "git_tag"
                ].initial = f"{repo.release_tag_regex}{next_version}"
```

(also includes a typo fix intial -> initial)

metaci/release/tests/test_admin.py
```diff
@@ -0,0 +1,21 @@
...
+def test_release_admin_init_from_repo():
+    release = ReleaseFactory(version_number="1.0")
+    repo = release.repo
+
+    admin = ReleaseAdmin(model=Release, admin_site=AdminSite())
+    request = RequestFactory().get(f"/?repo_id={repo.pk}")
+    form = admin.get_form(request)
+
+    assert form.base_fields["repo"].initial == repo
+    assert form.base_fields["version_number"].initial == "1.1"
```
davisagli, line 21: @jofsky I added a test

davisagli (approved):

PR #2246, opened 2021-05-12
Feature/robot 4

boakley:
This has some small changes to robot_importer to support changes in the output.xml schema for robot 4.x. I also converted all of the test xml files used by the robot importer tests. I did so by running robot's 'rebot' tool which has the ability to read the old format and write the new one. One test file also required a small tweak to work around a small bug (feature?) of rebot where it threw away an attribute that had an empty string as a value.

If you want to test this locally against robot 4, you'll need to upgrade your local version of CumulusCI to use robot 4 (see branch feature/robot_4) and then update `prod.in` to use the local version of CumulusCI (eg: replace `cumulusci` with  `-e /path/to/CumulusCI`)

Robot release notes for the output.xml changes can be found here: [Generated output.xml has been changed](https://github.com/robotframework/robotframework/blob/master/doc/releasenotes/rf-4.0.rst#generated-output-xml-has-been-changed)

### DIFF
--- a/metaci/testresults/robot_importer.py
+++ b/metaci/testresults/robot_importer.py
@@ -241,7 +241,7 @@ def find_screenshots(root):
     if root is None:
         return []
     screenshots = []
-    for msg in root.findall(".//msg[@html='yes']"):
+    for msg in root.findall(".//msg[@html='true']"):
         txt = "".join([text for text in msg.itertext()])
         for screenshot in re.findall(r'href="([\w.-]+)">', txt):
             screenshots.append(screenshot)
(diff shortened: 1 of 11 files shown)


boakley:
There's one heroku test failure that I'm not sure what to do about. That test passes on my local machine. Also, the tests on heroku are going to run with robot 3.x since it depends on CumulusCI. I'm not sure how I'm supposed to configure things so that the heroku tests use an unreleased version of CumulusCI.  Perhaps I can tweak the test to give us some more information than just "not found".

davisagli (approved):

PR #2374, opened 2021-06-16
Add option to call GUS for change traffic control

jofsky:


jofsky:
@davisagli This is for the most part done, just wanted to create a draft to discuss the error payloads I've been observing.

metaci/release/utils.py
```diff
@@ -220,14 +222,20 @@ def send_stop_webhook(release, role, config_item):
...
         json=payload,
         headers={"Authorization": f"Bearer {token}"},
     )
     result = response.json()
-    if result["success"]:
+    if (
+        "results" in result
+        and len(result["results"]) > 0
+        and result["results"][0]["success"]
+        or result["success"]
```
davisagli, line 234: The example success response shown at https://confluence.internal.salesforce.com/display/CHANGEMGMT/Change+APIs#expand-Start does not include a `success` key at the top level -- so I would expect this to raise KeyError

metaci/build/models.py
```diff
@@ -333,7 +333,7 @@ def run(self):
             # Look up or spin up the org
             org_config = self.get_org(project_config)
             if (
-                self.org and self.org.name and self.org.name.lower() == "packaging"
+                self.plan and self.plan.role and self.plan.role in ["release","release_deploy","push_sandbox","push_production"]
```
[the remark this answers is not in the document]
jofsky: This was intentional. I just wanted to ensure we always stopped plans. I agree I think a helper function is in order.

metaci/release/tests/test_models.py
```diff
@@ -30,3 +31,67 @@ def test_empty_release_init(self):
...
+    #                 ),
+    #                 plan=PlanFactory(),
+    #                 external_id="1000",
+    #                 start_time="2021-06-08T08:00:00+00:00",
+    #                 stop_time="2021-06-09T18:00:00+00:00",
+    #             ),
+    #         ],
+    #         bulk=False,
+    #     )
+    #     assert release.create_default_implementation_step("test") == None
```
jofsky, line 79: how can I test this without receiving  `ValueError: save() prohibited to prevent data loss due to unsaved related object 'release'.` I've attempted to use release.implementation_steps.set() with no success, any ideas @davisagli ?

davisagli (commented):
@jofsky oops I just realized I made some comments last week that I forgot to submit.

metaci/release/tests/test_models.py
davisagli: @jofsky can I see the stacktrace so I can see exactly where that error is happening?

metaci/build/models.py
```diff
@@ -333,13 +333,21 @@ def run(self):
             # Look up or spin up the org
             org_config = self.get_org(project_config)
             if (
-                self.org and self.org.name and self.org.name.lower() == "packaging"
+                self.plan.change_traffic_control
             ):  # Calling for any actions taken against packaging org
```
davisagli, line 337: This comment is no longer accurate

metaci/build/models.py
```diff
@@ -410,6 +418,27 @@ def run(self):
             )
             if org_config.created:
                 self.delete_org(org_config)
+
+            if (
+                self.plan.change_traffic_control
+            ):  # Calling for any actions taken against packaging org
```
davisagli, line 424: This comment is no longer accurate

metaci/build/models.py
```diff
@@ -410,6 +418,27 @@ def run(self):
...
+                    message = f"Error while sending implementation stop step webhook: {err}"
+                    self.logger.error(message)
+                    set_build_info(
+                        build, status="error", exception=message, time_end=timezone.now()
+                    )
+                    self.logger = init_logger(self)
+                    self.logger.error(str(e))
+                    self.delete_build_dir()
+                    self.flush_log()
+                    return
```
davisagli, line 441: Let's move this entire block for processing the webhook down between lines 444 and 445. That way:
a. We don't need to call delete_build_dir and flush_log from two different places
b. We'll be sure to log both the error that stopped the build and possibly an error while calling the webhook

Also let's remove set_build_info from this block. An error while sending the stop webhook should not overwrite the main error that stopped the build.

Don't we need to add a parameter to send_stop_webhook so that we can set the status on the implementation step differently based on whether the build succeeded or failed? We don't want it to end up as "Implemented - per plan" if it failed.

metaci/release/tests/test_models.py
```diff
@@ -30,3 +44,52 @@ def test_empty_release_init(self):
         assert not release.work_item_link  # checking default set to None
         assert release.change_case_template  # checking default see not None
         assert not release.change_case_link  # checking default set to None
+
+    ####### Keep getting the following error: ###############
+    # ValueError: save() prohibited to prevent data loss due to unsaved related object 'release'.
+    #########################################################
```
davisagli, line 50: Remove old comment

davisagli (changes requested):

metaci/release/models.py
```diff
@@ -123,37 +122,26 @@ def update_from_github(self):
...
+            start = datetime.datetime.combine(
+                self.release_creation_date
+                + datetime.timedelta(
+                    days=implementation_step.get("start_date_offset", 0)
+                ),
+                datetime.time(implementation_step["start_time"]),
+            )
+            end = start + datetime.timedelta(hours=implementation_step["duration"])
+            self.create_default_implementation_step(
+                implementation_step["plan"],
```
davisagli, line 135: Can we call this role instead of plan?

metaci/release/models.py
```diff
@@ -123,37 +122,26 @@ def update_from_github(self):
...
+                start,
+                end,
+            )
 
     def create_default_implementation_step(
-        self, role, start_date=None, end_date=None, start_time=None, stop_time=None
+        self,
+        role,
+        start=None,
+        end=None,
```
davisagli, line 144: start and end should be required and not default to None

metaci/release/utils.py
```diff
@@ -185,17 +191,24 @@ def send_start_webhook(release, role, config_item):
...
+        and result["results"][0]["success"]
+    ):
         logger.info(
             f"Successfully started implementation_step: {implementation_step_id}"
         )
+        return
     else:
-        raise Exception("\n".join(err["message"] for err in result["errors"]))
+        raise Exception(
+            "\n".join(err["message"] for err in result["results"]["errors"])
```
davisagli, line 210: This is not the format returned by gus-bus if there was an error: https://github.com/SFDO-Tooling/gus-bus/blob/main/gus_bus/main.py#L99

metaci/release/utils.py
```diff
@@ -220,14 +233,21 @@ def send_stop_webhook(release, role, config_item):
...
+        and result["results"][0]["success"]
+    ):
         logger.info(
             f"Successfully stopped implementation_step: {implementation_step_id}"
         )
+        return
     else:
-        raise Exception("\n".join(err["message"] for err in result["errors"]))
+        raise Exception(
+            "\n".join(err["message"] for err in result["results"]["errors"])
```
davisagli, line 252: This is not the format returned by gus-bus if there was an error: https://github.com/SFDO-Tooling/gus-bus/blob/main/gus_bus/main.py#L99

davisagli (changes requested):

metaci/build/models.py
```diff
@@ -333,13 +333,21 @@ def run(self):
...
+                        self.plan.role,
+                        self.org.configuration_item,
+                    )
+                except Exception as err:
+                    message = f"Error while sending implementation start step webhook: {err}"
+                    self.logger.error(message)
+                    set_build_info(
+                        build, status="error", exception=message, time_end=timezone.now()
+                    )
+                    return
```
davisagli, line 350: We don't need this exception handler; it will be handled the same way by the block starting on line 344.

Also you can simply write `if self.plan.change_traffic_control:` and not split it over 3 lines

metaci/build/models.py
```diff
@@ -410,9 +418,23 @@ def run(self):
...
+                try:
+                    send_stop_webhook(
+                        self.release,
+                        self.plan.role,
+                        self.org.configuration_item,
+                        "Failed - no impact",
+                    )
+                except Exception as err:
+                    message = f"Error while sending implementation stop step webhook: {err}"
+                    self.logger.error(message)
```
davisagli, line 437: Let's log the message from the error that was raised in `send_stop_webhook`:

```suggestion
                    self.logger.error(str(err))
```

metaci/build/models.py
```diff
@@ -410,9 +418,23 @@ def run(self):
             )
             if org_config.created:
                 self.delete_org(org_config)
+
             self.logger = init_logger(self)
             self.logger.error(str(e))
             self.delete_build_dir()
+            if (
+                self.plan.change_traffic_control
+            ):
```
davisagli, line 427:
```suggestion
            if self.plan.change_traffic_control:
```

metaci/build/models.py
```diff
@@ -425,20 +447,18 @@ def run(self):
         self.flush_log()
 
         if (
-            self.org and self.org.name and self.org.name.lower() == "packaging"
-        ):  # Calling for any actions taken against packaging org
+            self.plan.change_traffic_control
+        ):
```
davisagli, line 451:
```
if self.plan.change_traffic_control:
```

metaci/build/models.py
```diff
@@ -425,20 +447,18 @@ def run(self):
...
             try:
                 send_stop_webhook(
                     self.release,
                     self.plan.role,
                     self.org.configuration_item,
+                    "Implemented - per plan"
                 )
             except Exception as err:
                 message = f"Error while sending implementation stop step webhook: {err}"
                 self.logger.error(message)
```
davisagli, line 461:
```suggestion
                self.logger.error(str(err))
```

metaci/build/models.py
davisagli: Also this block isn't quite right. It needs to NOT return, because we still want the build status to get set to `success` at the end of the function. And it needs to go before `self.flush_log()` to make sure the log gets written.

metaci/release/models.py
```diff
@@ -48,6 +48,26 @@ def get_default_production_date():
...
+
+    def start(self, release, implementation_step):
+        if self and implementation_step:
+            return datetime.datetime.combine(
+                release.release_creation_date
+                + datetime.timedelta(
+                    days=implementation_step.get("start_date_offset", 0)
+                ),
+                datetime.time(implementation_step["start_time"]),
+            )
```
davisagli, line 65: There's no need to take implementation_step as a parameter for this method. You can use `self.start_date_offset` and `self.start_time`.

There's also no reason you need `if self and implementation_step:`

We should be sure to return a timezone-aware datetime (see https://docs.djangoproject.com/en/3.2/topics/i18n/timezones/ for background). You can do this by calling django.utils.timezone.make_aware(dt) where dt is the result from combine()

metaci/release/models.py
```diff
@@ -48,6 +48,26 @@ def get_default_production_date():
...
+
+    def end(self, start, implementation_step):
+        return start + datetime.timedelta(hours=implementation_step["duration"])
```
davisagli, line 68:
```suggestion
    def end(self, start):
        return start + datetime.timedelta(hours=self.duration)
```

metaci/release/models.py
```diff
@@ -123,37 +143,21 @@ def update_from_github(self):
...
-        )  # time will vary depending on product
+        for step_dict in self.repo.default_implementation_steps:
+            step_info = DefaultImplementationStep(**step_dict)
+            start = step_info.start(self, step_dict)
+            end = step_info.end(start, step_dict)
+            self.create_default_implementation_step(
+                step_dict["role"],
+                start,
+                end,
+            )
```
davisagli, line 154: You could just pass `step_info` here, and let `create_default_implementation_step` retrieve values from it.

metaci/release/utils.py
```diff
@@ -18,12 +19,20 @@ def implementation_payload(role, config_item, release):
...
+                )
+            )
+            .isoformat(),
+            "end_time": pytz.timezone("US/Pacific")
+            .localize(
+                release.implementation_steps.get(plan__role=role).stop_time.replace(
+                    tzinfo=None
+                )
+            )
+            .isoformat(),
```
davisagli, line 35: We should get the step from the release once to avoid repeating that query.

And I don't get what you're doing with the timezone. We can't assume US/Pacific. Based on my suggestion above in DefaultImplementationStep, start_time and stop_time should be coming to us as timezone-aware datetimes in the local timezone. Then we should be able to convert them to UTC by doing start_time.astimezone(None).isoformat()

metaci/release/utils.py
```diff
@@ -152,15 +161,12 @@ def send_submit_webhook(release, config_item=None):
...
         return
     else:
-        if "results" in result:
-            for error in result["results"]:
-                raise Exception("\n".join(err["message"] for err in error["errors"]))
-        else:
-            raise Exception("\n".join(err["message"] for err in result["errors"]))
+        raise Exception(
+            "\n".join(err["message"] for err in result["results"]["errors"])
+        )
```
davisagli, line 169: I don't think this is the right format returned by gus-bus.

metaci/release/utils.py
```diff
@@ -185,20 +191,26 @@ def send_start_webhook(release, role, config_item):
...
+        and len(result["results"]) > 0
+        and result["results"][0]["success"]
+    ):
         logger.info(
             f"Successfully started implementation_step: {implementation_step_id}"
         )
+        return
     else:
-        raise Exception("\n".join(err["message"] for err in result["errors"]))
+        msg = "\n".join(err["message"] for err in result["results"]["errors"])
```
davisagli, line 209: I don't think this is the right format returned by gus-bus.

metaci/release/utils.py
```diff
@@ -217,17 +229,24 @@ def send_stop_webhook(release, role, config_item):
...
+        and len(result["results"]) > 0
+        and result["results"][0]["success"]
+    ):
         logger.info(
             f"Successfully stopped implementation_step: {implementation_step_id}"
         )
+        return
     else:
-        raise Exception("\n".join(err["message"] for err in result["errors"]))
+        msg = "\n".join(err["message"] for err in result["results"]["errors"])
```
davisagli, line 251: I don't think this is the right format returned by gus-bus.

metaci/release/utils.py
jofsky: Do you mind showing me sorry I've tried this several times and didn't get expected behavior.

davisagli: @jofsky Can you show me what is happening when you do it the way I described above?

davisagli: @jofsky please make sure we only do `release.implementation_steps.get(plan__role=role)` once

davisagli (approved):

PR #2375, opened 2021-06-17
Serve files from S3 via Django rather than redirecting to S3

davisagli:
Result assets stored in S3 will now be downloaded and served by Django, instead of redirecting the browser to a temporary S3 URL. This is necessary because we're restricting the S3 buckets to only be accessible from MetaCI's private space.

Also makes sure that if a file is missing on S3 we serve a 404 response instead of 500.

### DIFF
--- a/metaci/testresults/views.py
+++ b/metaci/testresults/views.py
@@ -5,16 +5,17 @@ from tempfile import mkstemp
 
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import ObjectDoesNotExist
-from django.http import HttpResponse
-from django.shortcuts import get_object_or_404, redirect, render
+from django.http import Http404, HttpResponse
+from django.shortcuts import get_object_or_404, render
+from django.urls import reverse
 from django.views.decorators.clickjacking import xframe_options_exempt
 from robot import rebot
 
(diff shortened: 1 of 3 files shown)


metaci/testresults/views.py
```diff
@@ -350,4 +369,28 @@ def build_flow_compare_to(request, build_id, flow):
...
+def _serve_file_response(file, content_type):
+    try:
+        raise FileNotFoundError
+        response = HttpResponse(file, content_type=content_type)
+        response["Content-Disposition"] = (
+            "attachment; filename=%s" % file.name.split("/")[-1]
+        )
+        return response
+    except FileNotFoundError:
+        raise Http404("File not found.")
```
jstvz, line 396: Why not just `raise Http404` instead?

metaci/testresults/views.py
```diff
@@ -159,9 +161,26 @@ def resolve_asset_url(m):
         asset_id = int(m.group(2))
         try:
             asset = queryset.get(id=asset_id)
-            url = asset.asset.url
         except ObjectDoesNotExist:
```
jstvz, line 164: Thanks for sending me down the `foo = foo.filter(id=x).first() --> if foo:...` vs. `try...except` rabbit hole.

jstvz (approved):

PR #2377, opened 2021-06-18
Update setup docs

davisagli:
This updates both the "local" setup instructions in running.rst and the Docker image to fix various things that were out of date.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       - .env
     volumes:
       - ./docker/postgres:/var/lib/postgresql/data:delegated
-    image: postgres:9.6
+    image: postgres:12
     ports: 
     - "5432:5432/tcp"
     restart: always
@@ -19,7 +19,7 @@ services:
       context: .
       dockerfile: Dockerfile
     command: |
-        /bin/bash -c 'while !</dev/tcp/postgres/5432; do echo "Waiting for postgres databaseto start..."; /bin/sleep 1; done; \
+        /bin/bash -c 'while !</dev/tcp/postgres/5432; do echo "Waiting for postgres database to start..."; /bin/sleep 1; done; \
         /bin/sleep 1; /bin/sh /app/docker/utility/start_server.sh;'
     ports:
       - "8000:8000"
--- a/docker/utility/install_node.sh
+++ b/docker/utility/install_node.sh
@@ -30,6 +30,7 @@ case "${dpkgArch##*-}" in \
       4ED778F539E3634C779C87C6D7062848A1AB005C \
       A48C2BEE680E841632CD4E44F07496B3EB3C1762 \
       B9E2F5981AA6E0CD28160D9FF13993A75599653C \
+      C82FA3AE1CBEDC6BE46B9360C43CEC45C17AB93C \
   ; do \
       gpg --batch --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys "$key" || \
       gpg --batch --keyserver hkp://ipv4.pool.sks-keyservers.net --recv-keys "$key" || \
--- a/docs/running_docker.rst
+++ b/docs/running_docker.rst
@@ -41,20 +41,20 @@ Running MetaCI In Docker
 
 Below are the following steps necessary to run MetaCI on Docker:
 
-1. `.env File Creation and Variable Declaration`_
-    __ `.env File Creation and Variable Declaration`
+1. Configure MetaCI
+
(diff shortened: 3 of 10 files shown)


docker-compose.yml
```diff
@@ -30,3 +30,4 @@ services:
       - .env
     volumes:
       - .:/app
+      - /app/node_modules
```
davisagli, line 33: @jstvz what's this for?

docker/utility/install_node.sh
```diff
@@ -1,10 +1,13 @@
...
 # This file installs the given Node version to the docker image
-# dynamically changing download link based on environmental 
-# Architecture
 
-# NEED TO SET NODE_VERSION ENVIRONMENT VARIABLE if not set already
+# Check whether jq is availible, if so, read the engines.node value from package.json
+if command -v jq &> /dev/null
+then
+    NODE_VERSION=${NODE_VERSION:-$(jq --raw-output '.engines.node' package.json)}
+fi
```
davisagli, line 9: Nice.

docs/running_docker.rst
```diff
@@ -41,20 +41,20 @@ Running MetaCI In Docker
 
 Below are the following steps necessary to run MetaCI on Docker:
 
-1. `.env File Creation and Variable Declaration`_
-    __ `.env File Creation and Variable Declaration`
+1. Configure MetaCI
 
-2. `Building Your Docker Containers`_ 
-    __ `Building Your Docker Containers`
+   Copy ``.env.example`` to create a file named ``.env`` in the top level of the MetaCI repository.
```
davisagli, line 46: Part of the problem is probably that I forgot to add `.env.example` to my commit. I'll do that now...

config/settings/base.py
```diff
@@ -18,7 +18,8 @@
 APPS_DIR = ROOT_DIR.path("metaci")
 
 env = environ.Env()
-env.read_env()
+env_file_path = Path(ROOT_DIR, '.env')
+env.read_env(str(env_file_path))
```
davisagli, line 22: @jstvz why is this needed?

Outside of Docker, we tell people to copy env.example to env (no initial period) and source it before running things. That's needed to add our virtualenv and node bin to the path but also takes care of adding variables to the environment.

Inside Docker, docker-compose.yml sets the env_file to .env so it should take care of setting the environment variables.

davisagli (commented):
@jstvz Thanks for doing a pass on this and making it better. I added a couple questions about things I'm not quite following.

docker-compose.yml
jstvz: Because we're using a volume that mounts the local project in the web
image. The volume isn't mounted during the build, so node_modules wasn't showing up, which meant yarn couldn't find the commands to run `yarn serve`.

config/settings/base.py
jstvz: From 6f12711:

    Got the following warning:
    
    ```
    30 0.310 /usr/local/lib/python3.8/site-packages/environ/environ.py:628: UserWarning: /app/config/settings/.env doesn't exist - if you're not configuring your environment separately, create one.
    ```
    
    ...two reasons: 1) The container can't see the environment variables set
    by docker-compose until CMD runs, so it's not available in the build; 2)
    django-environ's logic for finding the project root relies on
    sys._getframe() and is broken (I think).
---
That said, I was pretty iffy about it, so I'm open to an alternate approach.

docker-compose.yml
davisagli: Ah, I get it now: this overlays /app/node_modules from the image on top of /app mounted from outside the container. We also need this for `sfdx`; I'll add it.

config/settings/base.py
davisagli: @jstvz It works for me without this when building with BUILD_ENV=dev (so we don't try to run the Django collectstatic command). I see the warning but it's not a problem (we aren't expecting the variables to come from django-environ reading this file).

For BUILD_ENV=production I do get an error due to the missing env vars, but I think we need to do more work on a solution so that we don't end up embedding secrets in the image. Either we need a way to collect static assets without loading Django settings, or we need a way to use secrets while building without keeping them in the image (see https://pythonspeed.com/articles/docker-build-secrets/). I'd like to come back to this later since my main goal in this PR is to fix local builds, and we aren't using the Docker image in production.

jstvz: I'm fine with backing this out then.

jstvz (approved):
This is working on a fresh checkout. Thanks.

PR #2487, opened 2021-07-28
🛢 [MetaCI] Updating default implementation default field

jofsky:


jofsky:
Any ideas what I'm doing wrong @davisagli ? Should i just put dict?

metaci/repository/migrations/0009_default_implementation_steps.py
```diff
@@ -13,6 +13,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name="repository",
             name="default_implementation_steps",
-            field=models.JSONField(default=[], null=True),
+            field=models.JSONField(default=list, null=True),
```
davisagli, line 16: Was this migration already run in production? If so then you need to create a new migration rather than changing the existing one.

metaci/repository/models.py
```diff
@@ -34,7 +34,7 @@ class Repository(models.Model):
     github_id = models.IntegerField(null=True, blank=True)
     url = models.URLField(max_length=255)
     release_tag_regex = models.CharField(max_length=255, blank=True, null=True)
-    default_implementation_steps = models.JSONField(null=True, default=[])
+    default_implementation_steps = models.JSONField(null=True, default=list)
```
davisagli, line 37: I think you also need to add blank=True here so that an empty list is accepted as valid input.

davisagli (changes requested):

davisagli (approved):

PR #2488, opened 2021-07-28
[MetaCI] Inject push job start times dynamically from implementation steps

jofsky:


metaci/repository/models.py
```diff
@@ -34,7 +34,7 @@ class Repository(models.Model):
     github_id = models.IntegerField(null=True, blank=True)
     url = models.URLField(max_length=255)
     release_tag_regex = models.CharField(max_length=255, blank=True, null=True)
-    default_implementation_steps = models.JSONField(null=True, default=[])
+    default_implementation_steps = models.JSONField(null=True, default=list)
```
davisagli, line 37: These last 2 files duplicate the changes in #2487; were you intending to submit them there or here?

davisagli (commented):
What's the reason for this? Is it to make sure the push gets scheduled immediately rather than a short time in the future?

jstvz:
@jofsky I would like to suggest that we fix the push task upstream so that it handles ISO-8601 correctly (see [W-9634243](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000RpEPYA0/view)), before we add the mis-formatted datetime here.

jstvz:
> @jofsky I would like to suggest that we fix the push task upstream so that it handles ISO-8601 correctly (see [W-9634243](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000RpEPYA0/view)), before we add the mis-formatted datetime here.

PR: SFDO-Tooling/CumulusCI#2769

metaci/build/models.py
```diff
@@ -704,16 +704,23 @@ def _get_flow_options(self) -> dict:
...
-            options["push_sandbox"] = {
-                "version": f"{self.build.release.version_number}",
-            }
+            try:
+                options["push_sandbox"] = {
+                    "version": f"{self.build.release.version_number}",
+                    "start_time": self.build.release.implementation_steps.get(plan__id=self.build.plan.id).push_time.strftime("%Y-%m-%dT%H:%M")
+                }
+            except Exception:
+                raise Exception(f"Ensure that you have declared an implementation step with the {self.build.plan.role}.")
```
davisagli, line 713: Since we're filtering by plan id, we should probably mention the expected plan name rather than a role?

metaci/build/models.py
```diff
@@ -704,16 +704,23 @@ def _get_flow_options(self) -> dict:
...
-                "version": f"{self.build.release.version_number}",
-            }
-
+            try:
+                options["push_all"] = {
+                    "version": f"{self.build.release.version_number}",
+                    "start_time": self.build.release.implementation_steps.get(plan__id=self.build.plan.id).push_time.strftime("%Y-%m-%dT%H:%M")
+                }
+            except Exception:
+                raise Exception(f"Ensure that you have declared an implementation step with the {self.build.plan.role}.")
```
davisagli, line 723: Ditto

metaci/build/tests/test_models.py
```diff
@@ -165,21 +165,46 @@ def test_get_flow_options_release(self):
...
+    #     build_flow.build.plan.role = "push_production"
+        
+    #     build_flow.build.release = Release(repo=RepositoryFactory())
+    #     build_flow.build.release.plan.id = 22
+    #     build_flow.build.release.repo.default_implementation_steps = {"start_date_offset": 1,"start_time": 13,"push_time": 21,"duration": 5, "role": "push_production"}
+    #     build_flow.build.release.repo.save()
+    #     build_flow.build.release.version_number = "1.0"
+    #     build_flow.build.release.save()
+    #     options = build_flow._get_flow_options()
+    #     assert options["push_all"]["version"] == "1.0"
```
davisagli, line 198: Can you update or remove these commented tests?

metaci/repository/migrations/0009_default_implementation_steps.py
```diff
@@ -13,6 +13,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name="repository",
             name="default_implementation_steps",
-            field=models.JSONField(default=[], null=True),
+            field=models.JSONField(default=list, null=True),
```
davisagli, line 16: Why are you modifying an existing migration that is not related to the changes in this PR?

davisagli (changes requested):

metaci/build/models.py
```diff
@@ -704,16 +704,23 @@ def _get_flow_options(self) -> dict:
...
-            options["push_sandbox"] = {
-                "version": f"{self.build.release.version_number}",
-            }
+            try:
+                options["push_sandbox"] = {
+                    "version": f"{self.build.release.version_number}",
+                    "start_time": self.build.release.implementation_steps.get(plan__id=self.build.plan.id).push_time
+                }
+            except Exception:
+                raise Exception(f"Ensure that you have declared an implementation step the plan {self.build.plan.name}.")
```
davisagli, line 713:
```suggestion
                raise Exception(f"Ensure that you have declared an implementation step with the plan {self.build.plan.name}.")
```

metaci/release/models.py
```diff
@@ -51,6 +53,7 @@ def get_default_production_date():
 
 class DefaultImplementationStep(BaseModel):
     start_date_offset: int = 0
+    push_time: int
```
davisagli, line 56: This should have a default value of None so that we aren't forced to enter it for default implementation steps that aren't push upgrades.

metaci/release/models.py
```diff
@@ -67,6 +70,9 @@ def start(self, release):
     def end(self, start):
         return start + datetime.timedelta(hours=self.duration)
 
+    def _push_time(self,start):
+        return start + datetime.timedelta(hours=self.push_time)
```
davisagli, line 74: return None if push_time is None

metaci/build/tests/test_models.py
```diff
@@ -167,38 +166,40 @@ def test_get_flow_options_release(self):
...
-        build_flow.build.release = Release(
-            repo=build_flow.build.repo,
-            change_case_template=change_case_template,
-        )
-        build_flow.build.release.version_number = "1.0"
-        build_flow.build.release.save()
-        options = build_flow._get_flow_options()
-        assert options["push_sandbox"]["version"] == "1.0"
-        assert options["push_sandbox"]["start_time"] is None
+        with pytest.raises(Exception):
```
davisagli, line 169: This is way too careless. There could be an exception on the very first statement and the test would pass. This should go around only the statement where the exception is expected, and you should use `with pytest.raises(Exception, match="some text")` to make sure the exception is the one we expected and not something else.

davisagli (approved):

PR #2509, opened 2021-08-09
Commit Status builds with Branch regex

davidmreed:
# Changes

- The Plan object now has a `commit_status_regex` field. The `regex` field is tested against the branch for all trigger types, while `commit_status_regex` is tested against the commit status only. A migration moves existing values of `regex` to `commit_status_regex` for Plans with the trigger type `"status"`.


davidmreed:
Migration tested locally, but haven't run builds.

metaci/plan/migrations/0040_plan_commit_status_regex.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 3.1.13 on 2021-08-09 16:52
+
+from django.db import migrations, models
+
+
+def migrate_regexes(apps, schema_editor):
+    Plan = apps.get_model("plan", "Plan")
+    for plan in Plan.objects.all():
+        if plan.trigger == "status" and plan.regex:
```
davisagli, line 9: We can avoid loading all of the plans from the db:

```suggestion
    for plan in Plan.objects.filter(trigger="status").iterator():
        if plan.regex:
```

metaci/plan/migrations/0040_plan_commit_status_regex.py
```diff
@@ -0,0 +1,35 @@
...
+            model_name="plan",
+            name="commit_status_regex",
+            field=models.CharField(
+                blank=True,
+                help_text="For Plans that use the trigger Commit Status, run builds when the commit status matches this regex.",
+                max_length=255,
+                null=True,
+            ),
+        ),
+        migrations.RunPython(migrate_regexes),
```
davisagli, line 34: The lack of a backward migration to copy from commit_status_regex to regex means that we'll lose the regex values if we have to migrate backward (the framework will run the default no-op backward migration for RunPython, and then delete the field). It's unlikely that we'll need to do that in production. Perhaps more likely locally if you need to migrate backward before switching to a different branch. Figured it was worth considering at least.

metaci/plan/models.py
```diff
@@ -123,9 +129,13 @@ class Meta:
...
-        if self.trigger != "manual" and not self.regex:
+        if self.trigger not in ["manual", "status"] and not self.regex:
             raise ValidationError(
-                "Plans with a non-manual trigger type must also specify a regex."
+                "Plans with a trigger type other than Manual or Commit Status must also specify a regex to match tags or branches."
+            )
+        if self.trigger != "status" and self.commit_status_regex:
+            raise ValidationError(
+                "Only Plans with a Commit Status trigger may specify a Commit Status Regex."
             )
```
davisagli, line 139: Looks like it's currently possible to save a plan with trigger=status but no commit_status_regex, which we probably shouldn't allow.

metaci/plan/models.py
```diff
@@ -191,9 +201,19 @@ def check_github_event(self, event, payload):
             and self.trigger == "status"
             and payload["state"] == "success"
         ):
-            if not re.match(self.regex, payload["context"]):
+            if not re.match(self.commit_status_regex, payload["context"]):
```
davisagli, line 204: This will break if commit_status_regex is None; see my prior comment re validation.

metaci/plan/models.py
```diff
@@ -191,9 +201,19 @@ def check_github_event(self, event, payload):
...
 
+            # If we also have a branch regex filter, run it.
+            if self.regex:
+                if not payload["ref"].startswith("refs/heads/"):
+                    return run_build, commit, commit_message
+
+                branch = payload["ref"][11:]
+                # Check the branch against regex
+                if not re.match(self.regex, branch):
+                    return run_build, commit, commit_message
```
davisagli, line 215: Maybe put this in a helper function used by both the trigger="commit" and trigger="status" blocks, instead of duplicating?

davisagli (changes requested):

metaci/plan/migrations/0040_plan_commit_status_regex.py
davidmreed: Adding it's quick enough, so I've done so, although migrating backwards in production would be destructive if we had a plan with both regexes populated.

metaci/plan/models.py
davidmreed: I did, but then realized that the structure this is looking for doesn't actually match between the status and commit webhook payloads. Refactoring.

davisagli: Good catch

davisagli (approved):

PR #2525, opened 2021-08-12
Feature/test manager

jofsky:


metaci/build/flows.py
```diff
@@ -42,3 +46,6 @@ def post_task(self, step, result):
 
         if "robot_outputdir" in result.return_values:
             import_robot_test_results(flowtask, result.return_values["robot_outputdir"])
+
+        if settings.TEST_MANGER_ENABLED:
```
boakley, line 50: I'm thinking this needs to move inside the `if "robot_outputdir"` condition, since for the moment we're only supporting the export of robot results. I think eventually we may want to support apex tests too, but for now, we only want to call this function is if the flag is set and we have robot results.

boakley (changes requested):
It looks like we're missing tests to verify that `gus_bus_test_manager` was actually called after test results have been imported, and that it's only called if the flag is set and we have robot results.

metaci/build/flows.py
jofsky: Good point.

metaci/build/flows.py
```diff
@@ -42,3 +46,7 @@ def post_task(self, step, result):
 
         if "robot_outputdir" in result.return_values:
             import_robot_test_results(flowtask, result.return_values["robot_outputdir"])
+            if settings.RESULT_EXPORT_ENABLED:
+                export_robot_test_results(
+                    flowtask.build_flow.build, result.return_values["robot_outputdir"]
+                )
```
davisagli, line 52: If this is going to export all test results associated with a build, then we should call it once after the build is done, not after each robot task. Otherwise builds which include multiple robot tasks will end up reporting the same test results more than once.

metaci/build/tests/test_flows.py
```diff
@@ -117,10 +117,19 @@ def test_post_task__single_robot_task(get_spec):
...
     """Test for scenario where there are multiple Robot tasks defined
     in a single flow. We want to make sure that test results and related
     assets are created and associated with the correct related objects."""
+    mocker.patch(
+        "metaci.build.flows.settings",
+        METACI_RELEASE_WEBHOOK_URL="https://webhook",
+        METACI_RELEASE_WEBHOOK_ISSUER="MetaCI",
+        METACI_RELEASE_WEBHOOK_AUTH_KEY="test",
+        DJANGO_TIME_ZONE="US/Pacific",
+        GUS_BUS_OWNER_ID="00G",
```
davisagli, line 130: Why do the above settings need to be set for this test?

metaci/testresults/robot_importer.py
```diff
@@ -276,3 +280,46 @@ def render_robot_test_xml(root, test):
 
     test_xml = ET.tostring(testroot, encoding="unicode")
     return re.sub(r"sid=.*<", "sid=MASKED<", test_xml)
+
+
+def export_robot_test_results(build, results_dir: str) -> None:
+    logger = logging.getLogger(__name__)
+    if (
+        not settings.METACI_RELEASE_WEBHOOK_URL
+        or not settings.GUS_BUS_OWNER_ID
```
davisagli, line 289: gus bus owner id is not relevant to reporting test results

metaci/testresults/tests/test_robot_importer.py
```diff
@@ -316,3 +316,27 @@ def test_import_perf_results():
...
+
+    with temporary_dir() as output_dir:
+        copyfile(
+            TEST_ROBOT_OUTPUT_FILES / "robot_with_failures.xml",
+            Path(output_dir) / "output.xml",
+        )
+        robot_importer.import_robot_test_results(FlowTaskFactory(), output_dir)
+        assert (
+            robot_importer.export_robot_test_results(FlowTaskFactory(), output_dir)
+            is None
```
davisagli, line 341: We need a meaningful assertion here, like making sure the expected outgoing API request was actually made.

metaci/testresults/robot_importer.py
```diff
@@ -132,7 +134,18 @@ def import_robot_test_results(flowtask, results_dir: str) -> None:
...
+
             testresult.save()
+        results.append(
+            {
+                "name": f"{result['name']}",
+                "group": f"{result['suite']['name']}",
+                "status": f"{result['status'].capitalize()}",
+                "start_time": f"{result['start_time']}",
+                "end_time": f"{result['end_time']}",
+            }
```
boakley, line 146: We need to include result['robot_tags'] as in this data structure. I think we might also want the URL to the individual test result. gus-bus doesn't currently expect it, but it probably should.

metaci/testresults/robot_importer.py
```diff
@@ -294,23 +319,11 @@ def export_robot_test_results(flowtest) -> None:
         "build": {
```
boakley, line 319: We need to add the release name here before this makes it into production, but I haven't yet figured out where we're going to get it.

boakley, line 319: We need to include the org name - would that be `flowtask.build_flow.build.org.name`?

metaci/testresults/robot_importer.py
```diff
@@ -294,23 +319,11 @@ def export_robot_test_results(flowtest) -> None:
...
             "name": flowtest.build_flow.build.plan.name,
             "number": flowtest.id,
-            "url": "http://www.example.com",  # build.get_external_url() returns 'detail': [{'loc': ['body', 'build', 'url'], 'msg': 'URL host invalid, top level domain required', 'type': 'value_error.url.host'}] when sent to gus-bus,
+            "url": "https://www.example.com/",  # flowtest.build_flow.build.get_external_url(),  # returns 'detail': [{'loc': ['body', 'build', 'url'], 'msg': 'URL host invalid, top level domain required', 'type': 'value_error.url.host'}] when sent to gus-bus,
```
boakley, line 322: What do we need to do to fix this? I don't quite understand what the comment is telling me.

metaci/testresults/robot_importer.py
```diff
@@ -284,7 +309,7 @@ def render_robot_test_xml(root, test):
     return re.sub(r"sid=.*<", "sid=MASKED<", test_xml)
 
 
-def export_robot_test_results(flowtest) -> None:
+def export_robot_test_results(flowtest, test_results) -> None:
```
boakley, line 312: Is there a reason this is `flowtest` instead of `flowtask`, or is that just a typo?

metaci/testresults/robot_importer.py
```diff
@@ -186,8 +197,20 @@ def parse_test(test, suite, root):
...
     setup_time = _robot_duration(setup.find("status")) if setup else zero
     teardown_time = _robot_duration(teardown.find("status")) if teardown else zero
-
-    # I'm not 100% convinced this is what we want. It's great in the
+    start_time = datetime.now().astimezone(None).isoformat()
+    end_time = datetime.now().astimezone(None).isoformat()
+    if setup:
+        start_time = (
+            _parse_robot_time(setup.find("status").attrib["starttime"])
+            .astimezone(None)
```
boakley, line 205: 👍 I appreciate you doing the timezone dance here.

metaci/testresults/robot_importer.py
```diff
@@ -326,5 +339,5 @@ def export_robot_test_results(flowtest) -> None:
         )
         return
```
boakley, line 340: This isn't a critique on the code, it's just me thinking out loud:

`result` should have an "Accepted" or "Rejected" record for every test case. I'm thinking we should probably log the rejected items so we can answer the inevitable question "why isn't my test showing up in testmanager"? I think if we do a good enough job we should never have tests rejected, since they would typically be rejected only if  we pass invalid data to testmanager.

metaci/build/tests/test_flows.py
```diff
@@ -72,7 +74,11 @@ def test_post_task__result_has_exception(get_spec):
 
 
 @pytest.mark.django_db
-def test_post_task__single_robot_task(get_spec):
+def test_post_task__single_robot_task(mocker, get_spec):
+    mocker.patch(
+        "metaci.build.flows.settings",
+        RESULT_EXPORT_ENABLED=False,
```
davisagli, line 80: Why do you need this, given that the default is False?

metaci/build/tests/test_flows.py
```diff
@@ -198,6 +212,79 @@ def test_post_task__multiple_robot_tasks(get_spec):
...
+        "metaci.testresults.robot_importer.settings",
+        METACI_RELEASE_WEBHOOK_URL="https://webhook",
+        METACI_RELEASE_WEBHOOK_ISSUER="MetaCI",
+        METACI_RELEASE_WEBHOOK_AUTH_KEY="test",
+    )
+    mocker.patch(
+        "metaci.release.utils.settings",
+        METACI_RELEASE_WEBHOOK_ISSUER="https://webhook",
+        METACI_RELEASE_WEBHOOK_AUTH_KEY="12345",
+    )
```
davisagli, line 244: Does it work to patch django.conf.settings instead of patching settings objects in a bunch of different modules? I'm not sure if it will, but if there is a singleton settings object which gets initialized once, it might

metaci/build/tests/test_flows.py
```diff
@@ -198,6 +212,79 @@ def test_post_task__multiple_robot_tasks(get_spec):
...
+        assert (
+            1
+            == BuildFlowAsset.objects.filter(
+                category="robot-output", build_flow=build_flow
+            ).count()
+        )
+        assert 1 == TestResult.objects.all().count()
+        assert 0 == TestResultAsset.objects.all().count()
+        assert len(responses.calls) == 1
+        assert responses.calls[0].request.url == "https://webhook/test-results/"
```
davisagli, line 285: Should we assert something more specific about what we're sending there?

metaci/testresults/robot_importer.py
```diff
@@ -132,7 +134,18 @@ def import_robot_test_results(flowtask, results_dir: str) -> None:
                 testresult.robot_xml = testresult.robot_xml.replace(
                     f'"{screenshot}"', f'"buildflowasset://{asset_id}"'
                 )
+
             testresult.save()
+        results.append(
+            {
+                "name": f"{result['name']}",
```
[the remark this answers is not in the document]
davisagli: ++

metaci/testresults/robot_importer.py
```diff
@@ -180,8 +197,20 @@ def parse_test(test, suite, root):
...
+            _parse_robot_time(setup.find("status").attrib["starttime"])
+            .astimezone(None)
+            .isoformat()
+        )
+        end_time = (
+            _parse_robot_time(setup.find("status").attrib["endtime"])
+            .astimezone(None)
+            .isoformat()
+        )
+        # I'm not 100% convinced this is what we want. It's great in the
```
davisagli, line 213:
```suggestion
    # I'm not 100% convinced this is what we want. It's great in the
```

metaci/testresults/robot_importer.py
davisagli: looks like a typo to me.

also please put a docstring to remind us what this does

metaci/testresults/robot_importer.py
```diff
@@ -276,3 +307,37 @@ def render_robot_test_xml(root, test):
 
     test_xml = ET.tostring(testroot, encoding="unicode")
     return re.sub(r"sid=.*<", "sid=MASKED<", test_xml)
+
+
+def export_robot_test_results(flowtest, test_results) -> None:
+    if not settings.METACI_RELEASE_WEBHOOK_URL or not flowtest:
+        return  # should we better error handle this for individual case message error handling?
```
davisagli, line 314:
```suggestion
        return
```

Also, how would it get called without a flowtask?

metaci/build/tests/test_flows.py
davisagli: Also, if there are settings that should be set to a single value for all tests that is different from the non-test default, I would do that in https://github.com/SFDO-Tooling/MetaCI/blob/main/config/settings/test.py rather than using mocks.

metaci/testresults/robot_importer.py
jofsky: @davisagli I need help with this.

metaci/testresults/robot_importer.py
jofsky: good catch

metaci/testresults/robot_importer.py
davisagli: https://github.com/SFDO-Tooling/gus-bus/blob/1f1fdeb5bf9200bfacc1216746d9e9a53c653534/gus_bus/testmanager.py#L38 is validating that the `url` is a URL, and apparently Django is using a URL with a hostname that has no top-level domain and pydantic doesn't like it.

You can make the tests use a different hostname (example.com?) by setting SITE_URL in https://github.com/SFDO-Tooling/MetaCI/blob/main/config/settings/test.py

metaci/testresults/robot_importer.py
jofsky: Hehe never know where you're gonna be 😄

metaci/testresults/robot_importer.py
```diff
@@ -276,3 +307,37 @@ def render_robot_test_xml(root, test):
...
+        "tests": test_results,
+    }
+
+    token = jwt_for_webhook()
+    response = requests.post(
+        f"{settings.METACI_RELEASE_WEBHOOK_URL}/test-results/",
+        json=payload,
+        headers={"Authorization": f"Bearer {token}"},
+    )
+    result = response.json()
```
davisagli, line 334: Always do response.raise_for_status() before actually accessing the json -- to make sure we got an HTTP 200 response and not some HTTP error

metaci/build/flows.py
```diff
@@ -37,8 +41,10 @@ def post_task(self, step, result):
...
-
         flowtask.save()
-
         if "robot_outputdir" in result.return_values:
-            import_robot_test_results(flowtask, result.return_values["robot_outputdir"])
+            test_results = import_robot_test_results(
+                flowtask, result.return_values["robot_outputdir"]
+            )
+            if settings.RESULT_EXPORT_ENABLED:
+                export_robot_test_results(flowtask, test_results)
```
davisagli, line 50: We need to put this in a try/except block that logs exceptions but doesn't block continued execution of the flow.

metaci/testresults/robot_importer.py
davisagli: The `return` line here does nothing since this if/else block is the end of the function

metaci/testresults/tests/test_robot_importer.py
```diff
@@ -316,3 +318,119 @@ def test_import_perf_results():
...
+        "metaci.testresults.tests.test_robot_importer.settings",
+        METACI_RELEASE_WEBHOOK_URL="https://webhook",
+    )
+    mocker.patch(
+        "metaci.release.utils.settings",
+        METACI_RELEASE_WEBHOOK_ISSUER="https://webhook",
+        METACI_RELEASE_WEBHOOK_AUTH_KEY="12345",
+    )
+
+    with pytest.raises(Exception):
```
davisagli, line 391: Please don't do this. It will pass if there's _any_ exception while running the test, even if it's not the one you expected. Put the pytest.raises context manager around one specific statement that you expect to fail, and if it's a generic exception, make sure you also use `match=` to confirm it's the message you expected.

jofsky: Thank you I've been looking for exactly what you instructed 🙏

metaci/build/tests/test_flows.py
jofsky: So that I can test it locally having it set to True and still have my local tests pass, do you have an alternative @davisagli ?

metaci/build/tests/test_flows.py
jofsky: Do you have any suggestions?

metaci/testresults/robot_importer.py
jofsky: Yes but then I think you need to change the GUS-BUS model to move it out of the BuildMetadata and into the Build Pydantic model here https://github.com/SFDO-Tooling/gus-bus/blob/main/gus_bus/testmanager.py#L35

Let's address everything else then we can modify this, since it's such a small change. Lets just make sure we don't forget.

boakley: Good catch. Yes, you're right. Good catch.

boakley: FWIW, Jason said he had an idea about how to get the release by examining the PRs. He was going to try to provide some code to me soon.

config/settings/test.py
```diff
@@ -74,3 +74,9 @@
 
 METACI_LONG_RUNNING_BUILD_CONFIG = {"app_name": "test-app"}
 METACI_LONG_RUNNING_BUILD_CLASS = "metaci.build.autoscaling.LocalOneOffBuilder"
+
+METACI_RELEASE_WEBHOOK_AUTH_KEY = "12345"
+METACI_RELEASE_WEBHOOK_ISSUER = "MetaCI"
+METACI_RELEASE_WEBHOOK_URL = "https://webhook"
+
+SITE_URL = "https://webhook"
```
davisagli, line 82: This is confusing test data. SITE_URL is the URL of the MetaCI instance, not the webhook

metaci/build/tests/test_flows.py
```diff
@@ -214,33 +206,15 @@ def test_post_task__multiple_robot_task_test_results_disabled(get_spec, mocker):
 
 @responses.activate
 @pytest.mark.django_db
+@mock.patch("django.conf.settings")
```
davisagli, line 209: Replacing the entire settings object with a mock doesn't seem like a good idea.

metaci/testresults/robot_importer.py
davisagli: We can maybe hardcode as SFDO 234 for now and adjust later?

metaci/testresults/robot_importer.py
```diff
@@ -319,8 +319,9 @@ def export_robot_test_results(flowtask, test_results) -> None:
     payload = {
         "build": {
             "name": flowtask.build_flow.build.plan.name,
+            "org": flowtask.build_flow.build.org.name,
             "number": flowtask.id,
-            "url": "https://www.example.com/",  # flowtask.build_flow.build.get_external_url(),  # returns 'detail': [{'loc': ['body', 'build', 'url'], 'msg': 'URL host invalid, top level domain required', 'type': 'value_error.url.host'}] when sent to gus-bus,
+            "url": f"{flowtask.build_flow.build.get_external_url()}",  # if SITE_URL is set and still returning a 422 try setting an additional 127.0.0.1 domain in you /etc/hosts file
```
davisagli, line 324: Specifically, one that includes more than one part since gus-bus' validation is looking for a tld

metaci/build/tests/test_flows.py
davisagli: We should probably figure out whether we intend for the test suite to be run with or without your local "env" file loaded. If not, this isn't needed. If so, it would probably be better to have a session-scoped pytest fixture that resets the environment so that it doesn't leak into tests. We don't need to tackle that in this PR though.

jofsky:
@davisagli Any ideas on how to resolve, since all the tests are passing?

boakley (commented):
What was the reason for commenting out all of these tests?

metaci/testresults/robot_importer.py
davisagli: Based on the discussion yesterday, we want to include the branch name in the build metadata, and let gus-bus do the logic to determine the release from that.

metaci/testresults/tests/test_robot_importer.py
```diff
@@ -316,3 +321,196 @@ def test_import_perf_results():
...
+    )
+    with temporary_dir() as output_dir:
+        copyfile(
+            TEST_ROBOT_OUTPUT_FILES / "robot_with_failures.xml",
+            Path(output_dir) / "output.xml",
+        )
+        test_results = robot_importer.import_robot_test_results(flowtask, output_dir)
+        robot_importer.export_robot_test_results(flowtask, test_results)
+        response = json.loads(responses.calls[0].request.body)
+        assert response["build"]["metadata"] == {
```
davisagli, line 512:
```suggestion
        actual = json.loads(responses.calls[0].request.body)
        assert actual["build"]["metadata"] == {
```

davisagli (approved):

boakley (approved):

PR #2555, opened 2021-08-31
[MetaCI] Better payload error handling

jofsky:


metaci/testresults/tests/test_robot_importer.py
```diff
@@ -514,3 +514,26 @@ def test_gus_bus_payload_metadata():
...
             "key2": "hello, world",
         }
+
+
+@responses.activate
+@pytest.mark.django_db
+def test_gus_bus_bad_payload_metadata():
+    """Verify that all build-specific metadata is added to the payload"""
+    flowtask = FlowTaskFactory()
+    flowtask.build_flow.build.org = OrgFactory()
```
davisagli, line 524: You shouldn't need this line; we changed the BuildFactory so that it automatically gets created with an org: https://github.com/SFDO-Tooling/MetaCI/blob/9dbddc76bc85e68d08568e9695828f126bf5264c/metaci/fixtures/factories.py#L115

metaci/testresults/tests/test_robot_importer.py
```diff
@@ -514,3 +514,26 @@ def test_gus_bus_payload_metadata():
...
+    flowtask.build_flow.build.repo.metadata = {"key1": True, "key2": "hello, world"}
+    responses.add(
+        "POST",
+        f"{settings.METACI_RELEASE_WEBHOOK_URL}/test-results/",
```
davisagli, line 528: The route in gus-bus is defined without a trailing slash (https://github.com/SFDO-Tooling/gus-bus/blob/main/gus_bus/main.py#L244); we should be sure that MetaCI calls it without the slash or there's an extra redirect.

metaci/testresults/tests/test_robot_importer.py
```diff
@@ -514,3 +514,26 @@ def test_gus_bus_payload_metadata():
...
+        json={"success": True},
+        status=422,
```
davisagli, line 530: Don't use fake data in tests that isn't realistic. If we got a 422 from gus-bus then the body would have information about what was wrong in the input, not `{"success": True}`

metaci/testresults/tests/test_robot_importer.py
```diff
@@ -514,3 +514,26 @@ def test_gus_bus_payload_metadata():
...
+    )
+    with temporary_dir() as output_dir:
+        copyfile(
+            TEST_ROBOT_OUTPUT_FILES / "robot_with_failures.xml",
+            Path(output_dir) / "output.xml",
+        )
+        test_results = robot_importer.import_robot_test_results(flowtask, output_dir)
+        with pytest.raises(Exception):
```
davisagli, line 538: This should use match= to confirm that the exception message contained something from the gus-bus response.

metaci/testresults/tests/test_robot_importer.py
jofsky: E           requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.
E           
E           Request: 
E           - POST https://webhook/test-results/
E           
E           Available matches:
E           - POST https://webhook/test-results URL does not match

is happening when I remove it from the tests @davisagli

davisagli: @jofsky you also need to change it here https://github.com/SFDO-Tooling/MetaCI/blob/main/metaci/testresults/robot_importer.py#L333

davisagli (dismissed):

davisagli (approved):

PR #2573, opened 2021-09-01
[MetaCI] CTC release integration

jofsky:


.heroku/release.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash 
+set -e
+
+if [ -n "$CTC_URL" ] ; then
+     # Run migration surrounded by CTC start/stop
+     python .heroku/check_change_traffic_control.py
+else
+     # Simply run the migration
+     python manage.py migrate --noinput
+     python .heroku/check_change_traffic_control.py
```
davisagli, line 10: The point of the if/else is that we don't want to call check_change_traffic_control.py if there's no CTC_URL

.heroku/check_change_traffic_control.py
```diff
@@ -0,0 +1,87 @@
...
+    source_url = f"{settings.ctc_repo_url}/commit/{settings.heroku_slug_commit}"
+    logger.info("Checking change traffic control for {source_url}")
+
+    # find Case and start implementation step
+    result = call_ctc(
+        f"{settings.ctc_url}/case/match-and-start",
+        params={
+            "sm_business_name_id": settings.ctc_sm_business_name_id,
+            "source_url": source_url,
+            "heroku_app_name": settings.heroku_app_name,
```
davisagli, line 58: step_description

davisagli (changes requested):

.heroku/release.sh
jofsky: Sorry I did this for my local testing.

.heroku/check_change_traffic_control.py
jofsky, line 58:
```suggestion
            "step_description": settings.heroku_app_name,
```

.heroku/release.sh
jofsky, line 10:
```suggestion
```

davisagli:
I think this looks ready to test once we've deployed https://github.com/SFDO-Tooling/gus-bus/pull/102

.heroku/check_change_traffic_control.py
```diff
@@ -0,0 +1,87 @@
...
+import jwt
+import pydantic
+import requests
+
+logger = logging.getLogger()
+
+
+class CTCSettings(pydantic.BaseSettings):
+    ctc_url: str
+    ctc_auth_key: str
```
davisagli, line 16: Change this to use metaci_release_webhook_url and metaci_release_webhook_auth_key so we don't have to duplicate

jofsky: @davisagli not sure you I correctly interpreted what you wanted, let me know.

davisagli (approved):

PR #2601, opened 2021-09-08
Week 36 Updates

Br4nd0R:


requirements/prod.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
+# This file is autogenerated by pip-compile
```
davisagli, line 2: you might need to update pip-tools to avoid some of the inconsequential changes in here

davisagli (approved):

PR #2609, opened 2021-09-14
🛢 [MetaCI] Adding release cohort model

jofsky:


Br4nd0R:
This ligtm; seems like it's covering all AC mentioned in: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000UdnYYAS/view

The third AC does mention:

> A Release has its own Merge Freeze Start and Merge Freeze End fields which are optional overrides for the Release's Cohort dates.

Is anything needed to make these fields "override" release cohorts merge freeze values? Or is the precense of the field on the Release object all we're looking for here? cc @davisagli

jofsky:
@Br4nd0R I assumed the cci task we create to cover merge freeze will pass in these fields depending on their presence i.e. follow the following logic, if product specific merge freeze start and stop dates are used we will use those otherwise we will pass in the Cohort values as the default when no values are defined. @davisagli 

@Br4nd0R something similar to https://github.com/SFDO-Tooling/MetaCI/blob/main/metaci/build/models.py#L698

davisagli:
@Br4nd0R @jofsky when we discussed on Monday we decided that we don't actually need the ability to override the merge freeze start and stop for a particular Release. That came up while discussing the release cohort view, and I think we just neglected to come back to this story and remove that AC.

Br4nd0R (commented):
This lgtm. @davisagli - I'll let you give the final rubber stamping.

metaci/release/models.py
```diff
@@ -13,6 +13,26 @@
...
+    status_choices = [
+        ("None", "None"),
+        ("Active", "Active"),
+        ("Cancelled", "Cancelled"),
+        ("Planned", "Planned"),
+    ]
+    status = models.CharField(
+        max_length=9,
+        choices=status_choices,
+        default="None",
```
davisagli, line 27: When would we use status="None"? Should we remove that option and default to Planned?

metaci/release/models.py
```diff
@@ -135,6 +154,13 @@ class Release(StatusModel):
         null=False,
         blank=False,
     )
+    release_cohort = models.ForeignKey(
+        "release.ReleaseCohort",
+        on_delete=models.PROTECT,
+        null=True,
+        blank=False,
```
davisagli, line 161: I think blank also needs to be set to True so that it's possible to edit the Release in the admin without setting a release cohort.

davisagli (changes requested):

davisagli (approved):

PR #2788, opened 2021-11-11
Automate running release builds

bethbrains:


metaci/build/migrations/0019_build_planrepo.py
```diff
@@ -10,7 +10,7 @@ def set_planrepository(apps, schema_editor):
     Build = apps.get_model('build', 'Build')
     for build in Build.objects.all().iterator():
         try:
-            build.planrepo = build.plan.planrepository_set.get(repo = build.repo)
+            build.planrepo = build.plan.planrepos.get(repo = build.repo)
```
davisagli, line 13: do search and replace to make sure we caught all uses of planrepository_set

metaci/release/models.py
```diff
@@ -129,7 +135,14 @@ def _push_time(self, start):
...
 class Release(StatusModel):
-    STATUS = Choices("draft", "published", "hidden")
+    STATUS = Choices(
+        ("draft", _("Draft")),
+        ("failed", _("Failed")),
+        ("completed", _("Completed")),
+        ("inprogress", _("In Progress")),
+        ("waiting", _("Waiting")),
+        ("blocked", _("Blocked")),
+    )
```
davisagli, line 145: @davidmreed Are these the right choices? What needs to happen to existing values of published and hidden?

metaci/release/tasks.py
```diff
@@ -1,16 +1,149 @@
...
+    release.error_message = _(message)
+    release.status = Release.STATUS.failed
+    release.save()
+
+    release.release_cohort.error_message = _("One or more releases failed.")
+    release.release_cohort.status = Release.STATUS.failed
+    release.release_cohort.save()
+
+
+def _run_release_builds(release: Release):
```
davisagli, line 52: docstring

metaci/release/tasks.py
```diff
@@ -1,16 +1,149 @@
...
+    FAILED_STATUSES = [BUILD_STATUSES.error, BUILD_STATUSES.fail]
+    COMPLETED_STATUSES = [BUILD_STATUSES.success, *FAILED_STATUSES]
+
+    # NOTE: bug where MetaCI builds can hang in In Progress forever
+    # will be painful here.
+    def running(builds: List[Build]) -> bool:
+        return bool(builds) and not all(b.status in COMPLETED_STATUSES for b in builds)
```
davisagli, line 59: Could be (but no strong opinion from me):
```suggestion
        return any(b.status not in COMPLETED_STATUSES for b in builds)
```

metaci/release/tasks.py
```diff
@@ -1,16 +1,149 @@
...
+
+    def last_failed(builds: List[Build]) -> bool:
+        return (
+            bool(builds)
+            and not running(builds)
+            and builds[-1].status in FAILED_STATUSES
+        )
+
+    def any_failed(builds: List[Build]) -> bool:
+        return bool(builds) and any(b.status in FAILED_STATUSES for b in builds)
```
davisagli, line 76:
```suggestion
        return any(b.status in FAILED_STATUSES for b in builds)
```

metaci/release/tasks.py
```diff
@@ -1,16 +1,149 @@
...
+    # Locate builds associated with this Release.
+    # Sort Builds so that we can find the latest of each Role.
+    # This supports multi-build retries (rather than rerunning a single Build).
+    # Note that our predicates (succeeded() et al.) check for running builds
+    # and return False if builds are still running, which is why we can
+    # reasonably use `time_end`.
+
+    builds = defaultdict(list)
+    for build in release.build_set.filter(
+        plan__role__in=["Upload Release", "Release Deploy", "Release Test"]
```
davisagli, line 92: Need to use the internal role names: https://github.com/SFDO-Tooling/MetaCI/blob/main/metaci/plan/models.py#L21

davisagli: (And similarly below)

metaci/release/tasks.py
```diff
@@ -1,16 +1,149 @@
...
+) -> QuerySet[PlanRepository]:
+    return release.repo.planrepos.filter(plan__role=role, active=True)
+
+
+def fail_release_with_message(release: Release, message: str):
+    release.error_message = _(message)
+    release.status = Release.STATUS.failed
+    release.save()
+
+    release.release_cohort.error_message = _("One or more releases failed.")
```
davisagli, line 47: release.release_cohort could be None

metaci/release/tasks.py
```diff
@@ -1,16 +1,149 @@
...
+    build.save()
+    if release.status != Release.STATUS.inprogress:
+        release.status = Release.STATUS.inprogress
+        release.save()
+
+
+def find_active_planrepos_by_role(
+    release: Release, role: str
+) -> QuerySet[PlanRepository]:
+    return release.repo.planrepos.filter(plan__role=role, active=True)
```
davisagli, line 39: I think we also need to filter for plan__active=True

davisagli: Might be able to do `release.repo.planrepos.should_run().filter(plan__role=role)`

metaci/release/models.py
davidmreed: Good callout, we're going to add a migration to map both removed statues to `completed` - the published and hidden distinction will be in MetaDeploy only.

davisagli: fwiw, there aren't actually any hidden releases in our production instance of MetaCI

davidmreed: Yep, always possible some third-party operator has some though.

davisagli: I don't think we ever made "hidden" actually do anything though?

metaci/build/migrations/0019_build_planrepo.py
davidmreed: This one's actually wrong, per our discussion earlier, @bethbrains - this is the other side of the "junction object" relationship. I'll push a fix.

davisagli: I'd advocate for changing the name on that side too so it's consistent, but understand if you don't want to add to the scope of this PR.

metaci/plan/templates/plan/detail.html
```diff
@@ -1,68 +1,98 @@
...
+      class="slds-text-title slds-truncate slds-m-bottom--xx-small"
+      title="Active?"
+    >
+      Active?
+    </p>
+    <p
+      class="
+        slds-text-body--regular
+        slds-truncate slds-badge
+        {%
```
davidmreed, line 73: The code formatter snarled this up good (they usually don't understand Django template tags). Let's revert all of the changes except for the `planrepo` swap.

davisagli: we could consider using a different extension (.html.tpl) for these so as not to provoke the formatter

metaci/release/models.py
```diff
@@ -16,16 +16,18 @@
...
-        ("Canceled", "Canceled"),
-        ("Completed", "Completed"),
-    ]
+    STATUS = Choices(
+        ("planned", _("Planned")),
+        ("approved", _("Approved")),
+        ("active", _("Active")),
+        ("canceled", _("Canceled")),
+        ("completed", _("Completed")),
+        ("failed", _("Failed")),
```
davidmreed, line 25: Need to add a migration because we changed the case of the API values.

metaci/build/migrations/0019_build_planrepo.py
davidmreed: Done.

metaci/release/models.py
davidmreed: Oh, huh. Django didn't notice this change and generate a migration at all, and we missed the `StatusModel` pattern when we implemented `ReleaseCohort` (which copies the use of `STATUS` but not the inheritance from `StatusModel`).

Given that it's a `CharField` under the hood, I guess it doesn't actually need a migration for the change to the `Choices`, just for mapping the old values.

davisagli (approved):
LGTM

PR #2791, opened 2021-12-01
Feature/release process automation  cohort execution

davidmreed:
# Changes

- Added functionality to execute builds for a release cohort in dependency order.


metaci/release/tests/test_tasks.py
```diff
@@ -420,3 +422,250 @@ def test_release_merge_freeze_if_safe__safe(smfs_mock):
...
+            GitHubDynamicDependency(github=left.repo.url),
+        ],
+        separate.repo.url: [],
+    }
+
+    def flatten_mock(self, context):
+        return flatten_results[self.github]
+
+    with unittest.mock.patch.object(GitHubDynamicDependency, "flatten", flatten_mock):
+        result = get_dependency_graph([top, left, right, separate])
```
bethbrains, line 630: Let's add a second call to `get_dependency_graph` where the cohort does not include `top` and assert that the returned graph is still the same.

bethbrains (changes requested):
one additional test case, also would still like @codexmarcel to take a look!

codexmarcel (approved):
LGTM, was fun working with @bethbrains @davidmreed on this!

Let's make sure we have a proper PR description prior to merging.

PR #2802, opened 2022-01-03
Remove the Perf (Beta) tab

davisagli:
and related code, including the React stack

Note that since there's no longer a js stack, I'm now using [Honcho](https://honcho.readthedocs.io/en/latest/index.html) to run the local Django server alongside the worker process. This means there are some changes to how local development works:
1. Honcho will automatically load environment variables from `.env`, so you no longer need to make sure environment variables are loaded before starting the server.
2. Run `make start` instead of `yarn serve` to start the app.
2. The app is served on port 5000 instead of 8080 & 8000.


CONTRIBUTING.rst
```diff
@@ -23,7 +23,7 @@ Running the Server
 Docker runs the server automatically. If you are not using Docker for 
 development, you can run it like this:
 
-- ``yarn serve``: starts development server (with watcher) at
+- ``make start``: starts development server (with watcher) at
   `<http://localhost:8080/>`_ (assets are served from ``dist/`` dir)
```
Br4nd0R, line 27:
```suggestion
  `<http://localhost:5000/>`_ (assets are served from ``dist/`` dir)
```

Br4nd0R: Also I believe assets are no longer present in a `dist` dir. They should be coming from `static/` dirs located in the root dir or each individual app dir for the Django project.

CONTRIBUTING.rst
```diff
@@ -66,7 +52,6 @@ automatically prepended to commit messages):
 - 📦 (``:package:``) -> ``pip install -r requirements/local.txt``
 - 🛢 (``:oil_drum:``) -> ``python manage.py migrate``
 - 🐈 (``:cat2:``) -> ``yarn``
```
Br4nd0R, line 54: Do we still need this line?

config/settings/base.py
```diff
@@ -230,7 +230,6 @@ def nl_separated_bytes_list(val: str) -> List[bytes]:
     str(APPS_DIR.path("static")),
     str(ROOT_DIR.path("dist")),
```
Br4nd0R, line 231: Will we still have static files in a `dist` directory? I thought webpack was what compiled and output things there?

config/settings/production.py
```diff
@@ -128,7 +128,6 @@
     str(APPS_DIR.path("static")),
     str(ROOT_DIR.path("dist", "prod")),
```
Br4nd0R, line 129: Same question here; will we still have stuff output to a `dist` directory w/o webpack?

pytest.ini
```diff
@@ -6,4 +6,4 @@ testpaths =
     metaci
 
 python_files = test_*.py
-norecursedirs = node_modules dist src
+norecursedirs = dist
```
Br4nd0R, line 9: Same comment as above. Still using `dist`?

Br4nd0R (commented):
I like all the red here :). I had questions around if we will still have `dist/` dir present or not. Searching the codebase also shows that it looks like there are still some references to port 8000/8080 in a few places. I can loop back and take a final look once the final changes are made.

Br4nd0R (approved):
Got MetaCI up-and-running on Docker.

PR #2806, opened 2022-01-18
Week 3 Dependency Updates

Br4nd0R:
# Notable Updates
* `redis` to `4.1.1` - Only breaking change mentioned in `4.*` was pertaining to [cluster_mode ](https://github.com/redis/redis-py/pull/1762), which I don't believe we're using.
* `faker` to `11.3.0` - No notable changes in [CHANGELOG](https://github.com/joke2k/faker/blob/master/CHANGELOG.md)

## Dependabot Alerts Resolved
* `pillow`
* `lxml`


requirements/prod.txt
```diff
@@ -199,22 +190,24 @@ markupsafe==2.0.1
 oauthlib==3.1.1
     # via requests-oauthlib
 packaging==21.3
-    # via bleach
-pillow==8.4.0
+    # via
+    #   bleach
+    #   redis
+pillow==9.0.0
```
prescod, line 196: This is a big jump. Anything to worry about?

davisagli: I suspect we aren't actually using Pillow

davisagli (approved):

PR #2809, opened 2022-01-28
Remove unneeded reference to cumulusci module that moved

prescod:
Remove an unneeded reference to a cumulusci module that changed names. After this change, MetaCI's test suite is compatible with both the old and new names.

### DIFF
--- a/metaci/users/management/commands/tests/test_packaging_org_users.py
+++ b/metaci/users/management/commands/tests/test_packaging_org_users.py
@@ -66,7 +66,6 @@ class TestPackagingOrgUsers(TestCase):
         OrgFactory(name="packaging", repo__name="RepoD")
         salesforce_client.query = fake_query
         c = Command()
-        print(sys.modules.keys())
         output = io.StringIO()
         c.handle(stream=output)
         assert json.loads(output.getvalue()) == {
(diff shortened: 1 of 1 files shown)


metaci/users/management/commands/tests/test_packaging_org_users.py
```diff
@@ -66,8 +66,8 @@ def test_run_command(self, salesforce_client, refresh_oauth):
         OrgFactory(name="packaging", repo__name="RepoD")
         salesforce_client.query = fake_query
         c = Command()
-        OrgConfigModule = sys.modules["cumulusci.core.config.OrgConfig"]
-        with mock.patch.object(OrgConfigModule, "SKIP_REFRESH", True):
+        print(sys.modules.keys())
+        with mock.patch("cumulusci.core.config.org_config.SKIP_REFRESH", True):
```
davisagli, line 70: @prescod shouldn't this patch be unnecessary? SKIP_REFRESH is only used inside `refresh_oauth_token`, which is patched in a decorator above.

metaci/users/management/commands/tests/test_packaging_org_users.py
```diff
@@ -66,10 +66,9 @@ def test_run_command(self, salesforce_client, refresh_oauth):
         OrgFactory(name="packaging", repo__name="RepoD")
         salesforce_client.query = fake_query
         c = Command()
-        OrgConfigModule = sys.modules["cumulusci.core.config.OrgConfig"]
-        with mock.patch.object(OrgConfigModule, "SKIP_REFRESH", True):
-            output = io.StringIO()
-            c.handle(stream=output)
+        print(sys.modules.keys())
```
davisagli, line 69:
```suggestion
```

davisagli (approved):

PR #2811, opened 2022-02-01
Release Process Automation: MetaPush Integration

davidmreed:


metaci/release/tasks.py
```diff
@@ -310,6 +320,120 @@ def all_deps_satisfied(
...
+
+    for release in rc.releases.all():
+        if not release.repo.metapush_enabled:
+            continue
+
+        github_url = release.repo.url
+        # We should have exactly one succeeded Release (Upload Release) build for this Release
+        build = release.build_set.filter(
+            plan__role="release", status=BUILD_STATUSES.success
+        ).first()
```
bethbrains, line 366: If it should return exactly one, should we use `.get()` in a `try` and catch the `DoesNotExist` error? We aren't protecting here against finding >1 and just taking whichever is found first. Maybe it's adequately enforced elsewhere?

metaci/release/tasks.py
```diff
@@ -310,6 +320,120 @@ def all_deps_satisfied(
...
+        if not build:
+            raise DependencyGraphError(
+                f"Unable to find Build with role Release on {release}"
+            )
+
+        dep_graph_item = get_package_ids_from_build(build)
+        if not dep_graph_item:
+            raise DependencyGraphError("Unable to source package details from Build")
+        dep_graph.__root__.append(dep_graph_item)
```
bethbrains, line 375: is the explicit `__root__` necessary?

metaci/release/tasks.py
```diff
@@ -310,6 +320,120 @@ def all_deps_satisfied(
...
+@job
+def send_to_metapush(rc: ReleaseCohort):
+    _send_to_metapush(rc)
+
+
+def _send_to_metapush(rc: ReleaseCohort):
+    token = settings.METAPUSH_AUTHENTICATION_TOKEN
+    endpoint = settings.METAPUSH_ENDPOINT_URL
+
+    if endpoint and token:
```
bethbrains, line 408: can we flip this and put the "not configured" return up here to avoid the long indentation? #teamearlyreturn

bethbrains (changes requested):
Some questions and readability suggestions, nothing huge.

metaci/release/tasks.py
davidmreed: Yes

metaci/release/tasks.py
```diff
@@ -278,6 +284,10 @@ def execute_active_release_cohorts():
         rc.status = ReleaseCohort.STATUS.completed
         rc.save()
 
+        # Send this Release Cohort to MetaPush for push upgrades, if configured.
+        if is_metapush_enabled() and rc.send_to_metapush:
```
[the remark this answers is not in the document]
davidmreed: It means that the cohort is eligible to be sent to MetaPush on completion.

metaci/release/tasks.py
```diff
@@ -310,6 +321,125 @@ def all_deps_satisfied(
...
+]
+
+
+def get_package_ids_from_build(build: Build) -> Optional[DependencyGraphItem]:
+    """Traverse the Flow Tasks associated with this Build until we locate
+    a package upload task. Return a DependencyGraphItem with the package
+    details from its output, or None if not found."""
+    for build_flow in build.flows.all():
+        for flow_task in build_flow.tasks.all():
+            if flow_task.class_path in PACKAGE_UPLOAD_TASK_PATHS:
```
jstvz, line 336: Guessing a filter is more trouble than it's worth?

jstvz (commented):
Couple of questions...

metaci/release/metapush.py
```diff
@@ -0,0 +1,16 @@
...
+    AllPackageId: str  # 033
+    AllPackageVersionId: str  # 04t
+    Dependencies: List[str]  # List of 04ts
+
+
+class DependencyGraph(BaseModel):
+    __root__: List[DependencyGraphItem]
+
+    def __iter__(self):
+        return iter(self.__root__)
```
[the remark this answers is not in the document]
davidmreed: I'll buy it. I wish Pydantic has better ergonomics for non-dict models.

metaci/release/tasks.py
```diff
@@ -310,6 +321,125 @@ def all_deps_satisfied(
...
+
+
+def _send_to_metapush(rc: ReleaseCohort):
+    token = settings.METAPUSH_AUTHENTICATION_TOKEN
+    endpoint = settings.METAPUSH_ENDPOINT_URL
+
+    if not (endpoint and token):
+        rc.metapush_error = "MetaPush configuration is missing"
+        rc.save()
+        return
```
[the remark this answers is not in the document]
davidmreed: Yes, in a narrow case - if the env vars are changes between the enqueueing of and the execution of the RQ job.

metaci/release/tasks.py
davidmreed: I think I was going for deterministic behavior vis-a-vis iteration order. I don't feel strongly about it, though.

davidmreed: I don't remember what was in my head. A filter would be nicer here; let me see if I can tweak that without breaking anything else.

metaci/release/tasks.py
```diff
@@ -310,6 +321,125 @@ def all_deps_satisfied(
...
+                    "A dependency package version is missing because one or more repositories are not MetaPush-enabled."
+                )
+            dep_graph_item.Dependencies.append(package_version)
+
+    return dep_graph
+
+
+@job
+def send_to_metapush(rc: ReleaseCohort):
+    _send_to_metapush(rc)  # pragma: no cover
```
davisagli, line 404: It may be clearer as is, but you could write `send_to_metapush = job(_send_to_metapush)` after the definition of _send_to_metapush

metaci/release/tasks.py
davidmreed: Done.

metaci/release/tasks.py
```diff
@@ -310,6 +320,120 @@ def all_deps_satisfied(
...
+    token = settings.METAPUSH_AUTHENTICATION_TOKEN
+    endpoint = settings.METAPUSH_ENDPOINT_URL
+
+    if endpoint and token:
+        # Take our dependency graph (GitHub URL-based) and convert to
+        # MetaPush-style 04t/033 id based. MetaPush does not know
+        # about repositories - just packages.
+
+        try:
+            metapush_graph = convert_dependency_graph_to_metapush(rc)
```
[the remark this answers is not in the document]
davidmreed: Because `convert_dependency_graph_to_metapush()` has a type annotation, the LSP should be able to know the type of `metapush_graph`, so it can provide relevant suggestions and corrections.

metaci/release/tasks.py
```diff
@@ -310,6 +320,127 @@ def all_deps_satisfied(
...
+    """Convert the Release Cohort's dependency graph (expressed in terms of GitHub URLs)
+    to a MetaPush-style dependency graph (expressed in terms of 04t and 033 package ids)
+    """
+    package_versions_to_github_urls = {}
+    github_urls_to_package_versions = {}
+    dep_graph = DependencyGraph.parse_obj([])
+
+    for release in rc.releases.all():
+        if not release.repo.metapush_enabled:
+            continue
```
jstvz, line 365: Why not fail right away instead of on line 398?

metaci/release/tasks.py
```diff
@@ -310,6 +320,127 @@ def all_deps_satisfied(
     )
 
 
+PACKAGE_UPLOAD_TASK_PATHS = [
+    "cumulusci.tasks.create_package_version.CreatePackageVersion",
+    "cumulusci.tasks.salesforce.PackageUpload",
+]
+
+
+def get_package_ids_from_build(build: Build) -> Optional[DependencyGraphItem]:
```
jstvz, line 329: 💭: Some of this logic can replace `release.utils.update_release_from_github()`, which is calling out to github to set  `release.package_version_id` (IMO, 033/04t ids should both be on the release object).

metaci/release/tasks.py
```diff
@@ -310,6 +320,127 @@ def all_deps_satisfied(
...
+    details from its output, or None if not found."""
+    flow_tasks = FlowTask.objects.filter(
+        build_flow__build=build,
+        class_path__in=PACKAGE_UPLOAD_TASK_PATHS
+    )
+    if flow_tasks.count() == 1:
+        flow_task = flow_tasks.first()
+        # Grab the return values, which will include the 033 and 04t ids
+        all_package_id = flow_task.return_values["package_id"]
+        # The upload tasks for 1GP and 2GP have different return_values shapes
```
jstvz, line 341: TODO: fix this upstream.

jstvz (approved):
This seems a touch too complicated, but maybe I'm missing some context.

bethbrains (approved):

PR #2814, opened 2022-02-08
[MetaCI] Handle Create Release Button Error for initial release creat…

jofsky:
…ion W-9668092

### DIFF
--- a/metaci/repository/models.py
+++ b/metaci/repository/models.py
@@ -68,8 +68,6 @@ class Repository(models.Model):
                 version_number=1.0,
                 git_tag=f"{self.release_tag_regex}1.0",
             )
-        except Repository.DoesNotExist:
-            return None
 
 
 class BranchManager(SoftDeletableManager):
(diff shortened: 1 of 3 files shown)


metaci/repository/models.py
```diff
@@ -57,10 +57,17 @@ def get_github_api(self):
...
-        except Repository.DoesNotExist:
-            return None
+            return self.releases.latest()  # Also for MetaCI create release button.
+        except Release.DoesNotExist:
+            Release(
+                repo=self,
+                version_name="1.0",
+                version_number=1.0,
+                git_tag=f"{self.release_tag_regex}1.0",
+            )
```
davisagli, line 70: @jofsky I'm not sure what you're trying to do here. You're instantiating a Release without saving it or returning it. I think you could leave it as `return None` and the important part of the fix is changing from Repository.DoesNotExist to Release.DoesNotExist, right?

davidmreed (approved):

PR #2834, opened 2022-06-09
@W-11203493 Playwright

jofsky:


jofsky:
Might need to adjust buildpacks for app.json for heroku-ci

boakley (commented):
@jofsky will this block of code _only_ run when creating a new buildpack, or will it run when we try to install MetaCI locally or in a docker image?

boakley (approved):
Looks good.

PR #2847, opened 2025-05-08
Fix chrome test after buildpack update.

cwen-sfdc:
To address release issue
https://dashboard.heroku.com/apps/metaci-stg/activity/releases/8f88ba95-4267-4bac-a755-d4e0f2c1045c
"Unable to open chrome"

### DIFF
--- a/.heroku/chrome-smoke-test.py
+++ b/.heroku/chrome-smoke-test.py
@@ -1,8 +1,10 @@
 import selenium
 import selenium.webdriver
-import sys
+import sys,os
 options = selenium.webdriver.chrome.options.Options()
 options.add_argument('--no-sandbox')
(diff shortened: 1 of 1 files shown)


.heroku/chrome-smoke-test.py
```diff
@@ -1,8 +1,10 @@
 import selenium
 import selenium.webdriver
-import sys
+import sys,os
```
vsbharath, line 3: we dont need `os` here

vsbharath, line 3:
```suggestion
import sys
```

cwen-sfdc: yes, removed debug code.

vsbharath (approved):

PR #2848, opened 2025-06-06
@W-18654372-enable-snyk-scan

anjalimishra30:


cwen-sfdc:
@anjalimishra30, is this .strata.yml used for SFCI Managed V2 pipeline?
Did this org SFDO-Tooling onboard to SFCI Managed V2?

